PR #181, opened 2026-06-19
fix(cire/web): point guest site at the live wedding slug (PUBLIC_WEDDING_SLUG)

chav-aniket:
## Root cause
The guest site (`cireweddings.com`) is a **single-wedding** Astro build: `index.astro` bakes `GET /api/invite/<PUBLIC_WEDDING_SLUG>` at build time. But `deploy.yml`'s `Build cire/web` step never injected `PUBLIC_WEDDING_SLUG`, so the build fell back to the placeholder `'cire-wedding'` — which **404s** in prod. Result: the public invite rendered pure defaults (no hero image, no theme, no copy) **no matter what the organiser saved in the builder**. This is why "nothing in the invite builder had any effect," and why the hero/theme work looked broken.

## Fix
Inject `PUBLIC_WEDDING_SLUG` from a repo Variable into the cire/web build env (same pattern as `PUBLIC_TURNSTILE_SITEKEY` / `PUBLIC_GOOGLE_MAPS_EMBED_KEY`).

## Required alongside this
Set repo Variable `PUBLIC_WEDDING_SLUG = vaishnavi-rox-5ecbe9` (the live wedding's slug), then the next deploy rebuilds the guest site pointed at the real wedding.


chav-aniket:
Superseded — going with SSR + path-based wedding resolution (no build-time slug var) per the chosen approach.

PR #369, opened 2026-08-02
Rename the live wedding, and stop the bare domain naming a wedding

chav-aniket:
## Summary

- The live wedding was renamed **Vaishnavi & Rox → Roxon & Vaishnavi**, slug `vaishnavi-rox-5ecbe9` → `roxon-vaishnavi-5ecbe9`. The rename itself was a **production D1 data change** already applied — cire is multi-tenant and holds no per-couple defaults in code, so what is in this diff is the stale references that change left behind (test fixtures, and comments naming the couple whose comp entitlements they describe). `hero_title` was already `R & V` and needed no change.
- The guest site's bare domain (`/`) used to resolve the **most-recently-created** wedding via a public `GET /api/primary-wedding` and 302 to `/<slug>`. On a multi-tenant product that served one arbitrary couple's invite to every visitor at the root, and let any anonymous caller learn whose invite was newest. `/` now 302s off-origin to the marketing site, and the "primary wedding" concept is deleted rather than re-pointed.
- A defect introduced by that second change — an empty `PUBLIC_MARKETING_URL` turning `/` into a redirect loop — was caught by the pre-merge reviews and fixed, with the contract now covered by tests.

Guests are unaffected: they have always arrived on their own `/<slug>` invite link, which is unchanged.

## Workspaces affected

- `@cire/web` — the bare-domain route, the URL resolver, env plumbing, tests
- `@cire/api` — deleted route + service method, one regression test
- `@cire/db` — schema comment only (index retention rationale)
- `@cire/organiser` — test fixtures only
- CI/infra: `.github/workflows/deploy.yml` (adds `PUBLIC_MARKETING_URL`)

## Decisions & issues

**[S-L1 / T-U1]** — Empty `PUBLIC_MARKETING_URL` turned `/` into a redirect loop
- **Issue:** `MARKETING_URL` resolved with `??`, which falls back only on `undefined`. The same commit shipped `.env.example` with `PUBLIC_MARKETING_URL=` (present-but-empty), so the documented setup path — copy the example to `.env` — inlined `""`. `Astro.redirect("", 302)` emits an empty `Location`, which resolves against the request URL, so `/` redirects to itself until the browser gives up.
- **Why:** It takes the guest site's root down in exactly the environment seeded from the repo's own example file. Production was unaffected (`deploy.yml` sets the value explicitly), but the changeset advertised the new test as the guard when that test only covered the *unset* case — the one path that already worked. Not an open redirect: an empty `Location` can only ever resolve same-origin, so no attacker-controlled destination is reachable.
- **Solution:** `resolveMarketingUrl(raw)` treats empty, whitespace, relative and non-`http(s)` values as unusable and falls back to the apex. It takes the raw value as an argument rather than reading `import.meta.env` internally. `.env.example` now ships the line commented rather than set empty.
- **Rationale:** The env var is Vite-inlined at build time, so a module-level constant is fixed at import and no test can vary it — as a function every branch is reachable. This mirrors `apiPreconnectHref` in the same directory, so it is the established local pattern rather than a new one. Verified against the failing input directly: the suite now passes with the value set empty and set relative, both of which broke before.

**[approach]** — Deleting the "primary wedding" concept rather than re-pointing it
- **Issue:** `GET /api/primary-wedding` was public and unauthenticated, and was the only **unscoped** read of the `weddings` table (`ORDER BY created_at DESC LIMIT 1`, no `WHERE`).
- **Why:** A single-tenant assumption held over from the bespoke era. With two weddings live it leaked which couple signed up most recently and served their invite at the root.
- **Solution:** Deleted the route, its tests, `weddingsService.primaryWeddingSlug()`, and the guest-side `fetchPrimaryWedding()`. `/` redirects to a build-time marketing URL with no API call.
- **Rationale:** There is no *correct* single wedding to resolve, so a better "primary" does not exist to pick. The security review confirmed every surviving `from(weddings)` is scoped by `id`, `slug` or `ownerOsnProfileId`, and no remaining unauthenticated route can be asked to enumerate weddings.

**[approach]** — Dropping the query string on the redirect
- **Issue:** The old route forwarded `?code=` across the redirect; the new one drops it.
- **Why:** `?code=` is a host-preview **claim credential**. Forwarding it off-origin would deposit it in the marketing site's access logs and `Referer` chain, and the `no-referrer` meta that protects it covers only the invite documents, not a cross-origin 302.
- **Solution:** `index.astro` reads nothing from `Astro.url` / `Astro.request`; a source-text test pins that.
- **Rationale:** The security review traced every producer and consumer of `?code=` — all build and read the path-routed `/<slug>?code=…` form. Nothing constructs a bare-domain `/?code=…`, so the forwarding was vestigial.

**[approach]** — 302 rather than 301
- **Issue:** A permanent redirect would be the more "correct" HTTP semantic for a fixed destination.
- **Why:** 301s are cached by browsers indefinitely and cannot be recalled.
- **Solution:** 302, with a test asserting `301` does not appear in the route.
- **Rationale:** Where the bare domain points is a product decision we expect to revisit; the cost of being wrong with a 301 is unbounded and unfixable, and the cost of a 302 is one Worker invocation.

**[P-I1]** — The bare-domain 302 is an uncacheable Worker invocation — **left open, not fixed**
- **Issue:** `index.astro` is an SSR route with no `Cache-Control`, so every hit on `/` wakes the Worker to emit a response that is byte-identical for a whole deploy. A 302 with no freshness directive is not heuristically cacheable per RFC 9111.
- **Why:** It spends an invocation from the account-wide 100k/day Free-tier budget shared with actual invite rendering. Bare-domain traffic is low, so the absolute cost is small.
- **Solution:** Tracked in `cire/wiki/todo/perf.md`. Fix is either a short `Cache-Control: public, max-age=300` or moving the redirect to the asset layer.
- **Rationale:** Deferred deliberately — it is an optimisation that only *became available* on this branch (the destination was previously request-dependent, which is why the old endpoint was correctly `no-store`), and it is orthogonal to the correctness change here. Bundling it would mix a caching decision into a security fix.

**[decision]** — Keeping `weddings_created_at_idx`
- **Issue:** Migration 0053 added the index specifically for the deleted query, leaving it near-dead.
- **Why:** A dead index is write amplification on every insert and update.
- **Solution:** Kept, and flagged in `cire/db/src/schema.ts` as a removal candidate for the next time this schema changes.
- **Rationale:** The performance review established the amplification does not materialise: SQLite skips index maintenance on `UPDATE` when the indexed column is absent from the SET list, and `created_at` is written once at insert and never mutated. The only cost is one row per *wedding created*, among the rarest writes in the system — against which dropping it costs a production migration plus a schema-drift touchpoint in `db/setup.ts`.

**[T-M1 / T-S1 / T-S2]** — The changed behaviour was assertable nowhere
- **Issue:** Every decision above lives in two lines of `.astro` frontmatter that no test touched; `middleware.test.ts` still used the removed same-origin redirect as its fixture; and nothing asserted the deleted endpoint stays deleted.
- **Why:** Reinstating `${Astro.url.search}`, flipping to 301, or re-mounting the route during a merge or revert would all have passed green.
- **Solution:** New `pages/index.test.ts` source-text guard (302, destination is `MARKETING_URL` not a literal, nothing read from the request); the middleware fixture moved to an off-origin `Location`; `app.test.ts` asserts `GET /api/primary-wedding` 404s.
- **Rationale:** The route renders nothing, so there is no DOM to assert against — a source-text guard is the cheapest thing that can actually fail, and `gold-text-token.test.ts` already establishes that approach over `.astro` files in this workspace. A removal made for a disclosure reason deserves a contract, not just an absent file.

**[note]** — Local `main` is 43 commits stale
- **Issue:** `git diff main...HEAD` reports 300+ files across osn, pulse, zap and shared — none of it this branch's work.
- **Why:** The branch was cut from the newer `origin/main` tip while the local ref lagged.
- **Solution:** All review and diff work in this PR used `origin/main` as the base. No repo change.
- **Rationale:** Flagging it so the diff is read against the right base; a local `git checkout main && git pull --ff-only` avoids the next branch looking equally alarming.

## Test plan

- [ ] `bun run --cwd cire/api test` → 1488 pass (was 1487; +1 for the primary-wedding 404 guard)
- [ ] `bun run --cwd cire/web test` → 699 pass (was 694; +5 for the resolver table test and route guard)
- [ ] `bun run --cwd cire/organiser test` → 696 pass (unchanged; fixture-only edits)
- [ ] `bun run --cwd cire/web build` succeeds; the compiled index chunk is `Astro.redirect(MARKETING_URL, 302)` with no API call
- [ ] `PUBLIC_MARKETING_URL="" bunx vitest run src/lib/marketing-url.test.ts` passes — this is the case that regressed
- [ ] After deploy, confirm `invite.cireweddings.com/` lands on `cireweddings.com` and `invite.cireweddings.com/roxon-vaishnavi-5ecbe9` renders the invite
- [ ] Confirm `api.cireweddings.com/api/primary-wedding` returns 404

**Note on verification:** the sandbox blocks outbound traffic to `cireweddings.com` and the Astro dev server does not start there, so the live-URL checks above have **not** been run — the production rename was verified by D1 read-back, and the redirect by the compiled bundle. The last two boxes need a human.

Also worth knowing: `@cire/*` packages have no typecheck script, so `bun run check` does not cover them (pre-existing, tracked in `cire/wiki/todo/api.md`). Tests and the build are the type safety net here.

https://claude.ai/code/session_017o6Qry3MqLsEeprwofpoBh

---
_Generated by [Claude Code](https://claude.ai/code/session_017o6Qry3MqLsEeprwofpoBh)_


chav-aniket:
Merged `main` in (`3929873`) to clear the conflicts. No code conflicted — both were in `cire/wiki/todo/` shards where #367/#368 and this branch each prepended an entry to the same spot.

**`security.md`** — two independent review sections, neither superseding the other. Kept both. The repeated `S-L1` id is fine; ids are scoped per review section, as this shard already does elsewhere.

**`api.md`** — kept both new entries, and un-mangled a third. Flagging this because it is a change to content that belongs to #368, not to this PR:

> #368 appears to have clobbered the `GET /api/claim/session` entry's title line. On `main` that bullet runs straight from `…organiser 696 -> 700. See [[guest-event-editor]] §3.2, [[web]].` into `(claude/invite-code-gating-hints-g8nw0o, 2026-08-02) — a cire_session lasts 30 days…`, and the title itself is gone (grepping `main` for it returns nothing).

This branch still had the intact copy, and the glued tail matches it verbatim, so restoring it was a mechanical split rather than a reconstruction — no wording was invented. Doing it here because this merge is exactly where it was fixable. Happy to revert that hunk if you would rather it were handled on #368's side.

Test counts in the description are now stale — post-merge they are **cire/api 1494**, **cire/web 699**, **cire/organiser 700**. The api and organiser increases are #368's own new tests, not this branch's. Changesets still validate.

---

PR #737, opened 2026-08-20
docs(legal): make the published notices describe the product that exists

chav-aniket:
## Summary

An audit of all eight published legal pages against Australian, EU/UK and US
requirements, and the fixes for what it found. Not a redraft — every change below
is a defect.

The headline: **`cire/invites`'s privacy notice and terms were written for one
wedding run by two individuals.** They named that couple, published a personal
email as the privacy contact, and claimed the Privacy Act's personal, family or
household exemption — *"we run this site as individuals planning a personal
celebration, not as a business"*. That was true of the first wedding, and has
been wrong for every guest of every wedding since. The household exemption
(Privacy Act s 16, GDPR Art. 2(2)(c)) covers a couple running their own event; it
has never covered the platform underneath them.

## Workspaces affected

`@cire/invites`, `@cire/landing`, `@osn/landing`, `@pulse/landing`, and a new
`@shared/legal`. **Two** changesets — `@cire/*` are version-less, and
`scripts/validate-changesets.sh` rejects a file mixing them with versioned
packages.

## Issues

**Closes**

| Issue | What |
|---|---|
| — | None — these were found by this audit, not previously filed. |

**Opened**

| Issue | What |
|---|---|
| — | None yet. Three items below need a decision, not a patch, and are listed under Out of scope for filing once you've read them. |

### The notice now describes the split the code implements — `approach`

- **Issue** — `cire/landing` said Cire processes guest data *on the couple's
  behalf* (a pure processor). `cire/invites` said the couple was responsible
  **and** outside the Privacy Act. Between them, guest data had no one owing
  duties over it.
- **Why** — the two notices contradicted each other, and the product contradicts
  the tidier of the two: `RETENTION_AFTER_FINAL_EVENT_MS` in
  `cire/api/src/services/retention.ts` deletes guest data one year after the
  final event, for every wedding, and no organiser can change it.
- **Solution** — hosts decide what to ask and what to do with the answers; we
  hold the data for them; **and** we alone decide retention, security posture and
  telemetry. The notice says the third part out loud.
- **Rationale** — choosing a retention period is a controller decision. Claiming
  processor-only status while unilaterally setting one is the kind of claim that
  falls apart precisely when it matters.

### Placeholders resolve from one file, and the draft banner is derived — `approach`

- **Issue** — every page carried its own `{{LEGAL_ENTITY}}`, `{{CONTACT_EMAIL}}`,
[...]

PR #738, opened 2026-08-20
docs(legal): give the identity app and Pulse the notices they promised

chav-aniket:
## Summary

Stacked on #737. Three published notices point at legal pages that do not exist:

- `@osn/landing` — *"covers visitors to this site only"*, and *"the OSN identity
  service and each connected app publish their own, separate privacy notices"*.
- `@pulse/landing` — using the app *"is governed by the OSN privacy notice shown
  when you sign in"*.

None of those notices existed. Neither app carried a legal page, or a link to
one, while holding email addresses, passkeys, sessions, the social graph, OIDC
grants, events, RSVPs and location. Both now serve `/privacy` and `/terms`.

## Workspaces affected

`@osn/social`, `@pulse/web`, `@cire/landing`. Two changesets — `@cire/landing` is
version-less and cannot share a file with the versioned two.

## Issues

**Closes**

| Issue | What |
|---|---|
| — | None — found by the audit in #737. |

**Opened**

| Issue | What |
|---|---|
| — | None yet; the link-placement gap under Out of scope needs your call first. |

### The notices are derived from the data map, not drafted fresh — `approach`

- **Issue** — a privacy notice written from scratch agrees with the code on the
  day it ships and drifts from then on.
- **Solution** — every claim comes from `wiki/compliance/data-map.md`, which
  already carried the lawful basis and retention per field. Where it says
  `Art. 6(1)(f)`, the page says legitimate interest and offers the objection;
  where it says 30-day sliding, the page says 30 days from last use.
- **Rationale** — the data map is maintained as part of shipping features, so it
  is the thing least likely to be wrong. A file header on each page says this, so
  the next person updates the map and these together.

### Two balancing tests are now readable by the people they are about — `approach`

- **Issue** — both lived only in the data map's balancing notes.
- **Solution** —
  - Pulse's **share attribution** stores a *platform name* and nothing else: no
    third-party identifier, no cookie, no cross-site token, visible only to that
    one event's organiser, on a legitimate-interest basis the attendee can object
    to.
  - **Attending an event can reveal something protected about you** even when the
    host published it freely. `Art. 9(2)(e)` — manifestly made public — covers the
    *host*, who chose to publish. It does not cover the *attendee*, who did not.
    The notice says attendance at events of that kind is treated as needing
    consent.
- **Rationale** — a balancing test the data subject cannot read is not much of a
  balancing test.

### Both routes render signed-out — `approach`

- **Solution** — `/privacy` and `/terms` are outside any auth gate in both apps.
- **Rationale** — a notice you can only read once you have already handed over
  your data is not a notice.

### `cire/landing`'s scope sentence now names the vendor portal — `approach`

- **Issue** — it listed the marketing site, the invitation site and the organiser
  portal. The vendor portal is a fourth surface on the same service.
- **Rationale** — `cire/host` and `cire/vendor` genuinely are covered by that
  notice; they were missing from the sentence, not from the coverage.

**Out of scope**

- **Neither app has a footer**, so there is nowhere established to link these
  from. Both pages are reachable by URL and cross-link each other. Where a
  persistent link belongs is a UI decision, not a legal one, and inventing chrome
  inside a legal PR is how a legal PR stops being reviewable. Worth filing once
  you've decided — my suggestion would be the existing Settings page in each app
  plus the OIDC consent screen, which is exactly where someone is being asked to
  agree to something.
- `cire/host` and `cire/vendor` still carry no link to the cire notices that
  cover them, for the same reason.

## Test plan

| Gate | Result |
|---|---|
| `bun run check` | ✅ |
| `bun run lint` | ✅ |
| `bun run fmt:check` | ✅ 1357 files |
| `bun run test` | ✅ 33/33 packages |
| `bun run --cwd osn/social build` | ✅ |
| `bun run --cwd pulse/web build` | ✅ |

Not verified: nobody has loaded either route in a browser. They are new routes in
two SPAs and both builds pass, but the rendered look is unchecked — worth a glance
on the previews.

Not verified, and more important: **no lawyer has read any of this.** It is an
engineering pass that makes the published claims match what the code does. That is
a necessary condition for these being correct, not a sufficient one.

### DIFF
--- a/.changeset/legal-app-pages-cire.md
+++ b/.changeset/legal-app-pages-cire.md
@@ -1,0 +1,7 @@
+---
+"@cire/landing": patch
+---
+
+Name the vendor portal in the privacy notice's scope sentence. The notice already
+covered the marketing site, the invitation site and the organiser portal; the
+vendor portal is a fourth surface on the same service and was simply missed.
--- a/.changeset/legal-app-pages.md
+++ b/.changeset/legal-app-pages.md
@@ -1,0 +1,24 @@
+---
+"@osn/social": patch
+"@pulse/web": patch
+---
+
+Give the identity app and Pulse the privacy notices and terms they were already
+promising.
+
+`@osn/landing`'s notice says it "covers visitors to this site only" and that "the
+OSN identity service and each connected app publish their own, separate privacy
+notices". `@pulse/landing`'s says using the app "is governed by the OSN privacy
+notice shown when you sign in". None of those notices existed, and neither app
+carried a legal page or so much as a link to one — while holding accounts,
+passkeys, sessions, the social graph, OIDC grants, events, RSVPs and location.
+
+Both apps now serve `/privacy` and `/terms`, reachable without signing in, with
+every claim taken from `wiki/compliance/data-map.md` — the lawful basis per
+purpose, the real retention windows, and where the data is stored.
+
+Two things the notices say that were not said anywhere before: Pulse's share
+attribution records a platform name and nothing else, visible only to that event's
+organiser, and it rests on legitimate interest you can object to; and attending an
+event can reveal something protected about you even when the host published it
+freely, so that is treated as needing consent rather than as manifestly public.
--- a/bun.lock
+++ b/bun.lock
@@ -276,6 +276,7 @@
       "dependencies": {
         "@osn/client": "workspace:*",
         "@osn/ui": "workspace:*",
+        "@shared/legal": "workspace:*",
         "@simplewebauthn/browser": "^13.3.0",
         "@solidjs/router": "^0.16.2",
         "@tailwindcss/vite": "^4.3.3",
@@ -400,6 +401,7 @@
       "dependencies": {
         "@osn/ui": "workspace:*",
         "@pulse/api": "workspace:*",
+        "@shared/legal": "workspace:*",
         "@shared/rp-auth": "workspace:*",
         "@solidjs/meta": "^0.29.4",
         "@solidjs/router": "^0.16.2",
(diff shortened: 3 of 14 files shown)


chav-aniket:
Cold review by a second model. Two findings here, and the first is the worst one in either PR — closed in `0294193f`.

**The pulse notice claimed an Art. 9 consent flow that does not exist.** It said we *"treat attendance at events of that kind as protected information, **ask before recording it**"*. Nothing asks. The line came from a data-map row that is a **directive, not a description** — `data-map.md:84` says *"Treat with extra care: explicit consent banner on RSVP for events tagged with sensitive categories"* — and no such banner, tag or gate exists anywhere in pulse. Compare cire, where the same promise is real and enforced (`rsvp.ts`, 422 without `dietaryConsent`).

That is exactly the failure the old cire page had and this branch exists to end. The page now says what is true: the host's visibility setting and your choice whether to respond are the controls today, the separate ask doesn't exist yet, and it should. A test bans the old sentence returning before the feature does.

**"Pulse loads a map, and that is the only third-party content on it"** — two more, both firing from the browser:

- **OpenStreetMap** tile servers see the IP of anyone opening a page with a map.
- **Photon (Komoot)** receives every keystroke of an address plus the IP, as you type. `subprocessors.md` rates this **High** and says *"Block until proxied + consent banner added"* (S-M13). It appeared in no notice at all.

Both are now named, the Photon entry states plainly that it is not behind a gate and that we are moving it, and the page gains the recipients list it never had.

Also: withdrawing an OIDC grant does **not** delete it — the row is kept, marked withdrawn, until account deletion, because the row *is* the withdrawal record. The friendlier claim was the wrong one.

And the draft banner printed a repo path to the reader here too. Same leak, same fix.

Both apps gain tests mirroring the cire ones, plus the two claims above pinned so they can't quietly return.

PR #760, opened 2026-08-22
feat(cire-api): connect a couple's Stripe account (cash gifts, 1/n)

chav-aniket:
Stacked on #759.

## What

The first half of cash gifts: Stripe Connect onboarding and the account webhook.
The guest-facing "give money" surface and hosted Checkout follow on top of this.

**The couple are the merchant of record.** Gifts will be direct charges on their
own connected account, so the money goes from the guest's card to their bank and
cire never holds it. Holding gift funds, even in transit, is money transmission —
that is the whole reason the integration is Connect rather than a platform
checkout, and the schema was already shaped for it.

## Issues

**Closes** None. This is 1/n of the stack that delivers xchromo/osn#699 (registry cash gifts over Stripe Connect); the closing line sits on the top PR of the stack so the epic is not closed while the surface is half-built.

| Issue | Relation |
|---|---|
| xchromo/osn#699 | Epic this PR is the first part of — Express onboarding, the account-status route and the webhook skeleton |

**Opened** None.

## No `stripe` package

Three REST calls and a signature check over `fetch` (`services/stripe.ts`). The
official SDK is built around Node's http stack, and cire-api ships inside a 1MB
compressed Worker budget it already shares with Elysia, Drizzle and Effect. A
trade worth re-running if this file grows past what the gift flow needs — the
header says so.

## Routes

| Route | Gate |
|---|---|
| `POST …/registry/stripe/session` | `weddingOwner` + registry entitlement |
| `POST …/registry/stripe/refresh` | `weddingOwner` + registry entitlement |
| `POST /api/stripe/webhook` | Stripe signature (the only authentication it has) |

**Owner-only, not editor.** Every other registry write is `weddingEditor`,
because adding a gift is ordinary help. This names the bank account the money
lands in, and belongs on the same side of the role line as codes, deletion and
co-host removal.

**Create-or-resume, never create-again.** Onboarding is a form people abandon and
come back to. The route reuses any account already on the settings row;
`stripe_account_id` is written through a `coalesce` so a second create can only
fill a null, never repoint a couple's payouts; and Stripe's idempotency key is
the second belt against a double-tapped button.

**`refresh` is one live read**, for the single moment it matters: the couple have
just come back from onboarding and `account.updated` may be seconds behind them.
Everything else reads the cached columns.

## The webhook

Verified over the **raw request text** (a body parsed and re-serialised has
already lost the property being checked), compared without a length- or
value-dependent early exit, against **every** `v1` digest Stripe sent so a secret
can be rotated, and refused outside a **300-second window** — a valid signature
is valid forever, and without a window a captured delivery can be replayed at any
future point against the handler that records money.

A 200 does not mean "handled": an event type this product does not act on is
acknowledged, because the endpoint belongs to the platform account and a non-2xx
buys days of retries for something nobody was going to do anything with.

## Intent vs capability

`cash_gifts_enabled` is the couple's decision; `stripe_charges_enabled` is
Stripe's. The webhook writes only the second — clearing intent because a
capability lapsed would quietly turn the feature off for good, and restoring it
later would be us deciding something they never did. `PUT /registry/settings`
still refuses to enable cash gifts while Stripe cannot take a charge.

## It ships inert

Both secrets are key-optional and independent. No `STRIPE_SECRET_KEY` ⇒ the
onboarding routes are not mounted, so a deployment without a Stripe account has
no payment surface rather than a broken one. No `STRIPE_WEBHOOK_SECRET` ⇒ the
webhook route does not exist, because nothing else authenticates it. Documented
in `wrangler.toml`; nothing is reachable until someone sets them.

`/prep-pr` ran on this branch; five findings are fixed here rather than filed.

**S-H1 — the one that mattered.** `account.updated` was applied
unconditionally, and Stripe neither guarantees delivery order nor stops retrying
for three days. An older event carrying `charges_enabled: true`, redelivered
after Stripe had disabled the account, would re-enable it here — and that column
is the only gate on whether a couple may show guests a contribute button, so it
re-opens a payment surface Stripe has shut. The write now carries Stripe's own
`created` and refuses anything older than what the row already holds.

**S-L2**, the same shape one layer up: `attachStripeAccount` guarded the account
id with a `coalesce` but wrote the capability booleans unconditionally, so a
second call could leave the row describing account A's id beside account B's
capabilities. They now follow the id.

**S-M2.** The webhook is public, unauthenticated and unlimited — the signature
IS the authentication, and it cannot run until the body has been read, so the
body is what an attacker chooses the size of. It now rejects on a missing
signature header and on a declared `content-length` over 64KB before buffering
anything, and re-checks the materialised length because `Content-Length` is
theirs to lie about.

**S-M1.** Both onboarding routes spend an outbound Stripe call per request and
carried no limiter, so an owner holding the button burns the *platform's* Stripe
quota — a cross-tenant denial of service reached from one tenant's credentials.
They now carry a per-organiser limiter, after the role and entitlement gates, and
every Stripe call has a ten-second timeout.

**S-L1 / S-L3.** A revoked key, a withdrawn capability and a Stripe outage
produced an identical bare 502 with nothing to tell them apart — the status and
Stripe's own code are now logged, never its message. And if the row write fails
after Stripe created the account, the id is logged at the one moment it would
otherwise be lost: Stripe's idempotency key expires after 24 hours, so a later
retry would otherwise mint a second account for one couple.

## The paperwork, which is not done

`wiki/compliance/` now records the true state rather than an intention:

- **The Stripe DPA is unsigned.** Guest card-adjacent data reaches Stripe the
  moment this ships with a live key. Signing it is a **human merge-blocker on
  this PR** — nothing in the code can stand in for it.
- The Connect platform agreement, and whether Australia needs one executed
  separately, is an open question in the register.

## Tests

1746 pass. The client's form encoding, idempotency key, error-code-only failures
and network failure-as-value; the signature check against unsigned, mis-signed,
altered, stale and unconfigured deliveries; the routes' owner gate, entitlement
gate, create-or-resume, limiter, timeout, 502-not-500 on Stripe trouble; and the
webhook's monotonic guard, its size bounds and its leaving the couple's intent
alone.

Not verified against real Stripe — that needs a platform account, a test key and
a webhook endpoint. The code is written to be inert until they exist.

## Next in the stack

Hosted Checkout, `checkout.session.completed` → a `registry_contributions` row
with the balance-transaction FX capture, the organiser's connect panel, and the
guest-side "give money" surface — available whether or not anything is left on
the list.


chav-aniket:

---

PR #762, opened 2026-08-22
feat(cire): let a guest give money, list or no list (cash gifts, 2/n)

chav-aniket:
Stacked on #760 (which is stacked on #759).

## What

The other half of cash gifts: hosted Checkout, the gift the webhook writes when
Stripe says the money moved, and the panel a guest actually uses.

> Regardless of whether there's items left or not there should be the option of
> just giving money.

That is the shape: the panel sits above the shelves and **outside** the
items-exist branch, so it is in the same place whether the list is full, empty,
or entirely spoken for. A money option that appears only once the list runs dry
reads as a consolation prize.

## Issues

**Closes** None. This is 2/n of the stack that delivers xchromo/osn#699 (registry cash gifts over Stripe Connect); the closing line sits on the top PR of the stack so the epic is not closed while the surface is half-built.

| Issue | Relation |
|---|---|
| xchromo/osn#699 | Epic this PR is the second part of — the guest's give-money path, hosted Checkout, and the contribution record |

**Opened** None.

## The guest's side

`POST /api/invite/:slug/registry/contribute` answers with a Stripe Checkout URL
for a **direct charge** on the couple's connected account — the money is theirs
from the moment it is taken, and cire never holds it.

Every gate runs **before Stripe does**: the registry must be visible, the
household must belong to this wedding, the couple must have said yes, Stripe
must be able to take a charge today, and there must be an account to take it
into. A guest is turned away before their card is, or not at all. The
cash-specific refusal is its own 409 `cash_gifts_unavailable`; everything else
is the same 404 as ever.

Its own per-IP limiter, at the claim route's 20/min. Every call is an outbound
Stripe request — the same "amplifier" shape the link-preview route is limited
for, and the one guest route that can cost money to be wrong about — but the key
is an IP, not a household, and a NAT'd reception is one address for the whole
room. 5/min would have met the evening's sixth guest with a 429 on their way to
paying.

## The record

**The row is written first, as `pending`, before the guest is handed a payment
page** — and Stripe is told one opaque id and the money, nothing else.

That ordering is the whole design, and it came out of the review gate:

- **Stripe learns nothing about the guest.** The amount, the household, the note
  and the chosen name live in the pending row the id names. Sending them as
  session metadata was a transfer of guest personal data to a US processor that
  no register declared and that reconciliation never needed (C-H2).
- **A forged session settles nothing.** This endpoint also hears about sessions a
  connected account created for *itself*, where every metadata field is whatever
  its owner typed. When the row was written *from* the event, a couple could mint
  gift rows for money that never moved, with attacker-chosen text attributed to a
  named household (S-M1). Now the webhook settles the row that id names, after
  checking the session it just completed is the one on the row and that the row's
  wedding owns the account the event arrived on. There is no row to conjure.
- **No payment without a record.** If the row cannot be written, the guest is not
  sent to pay at all — a payment with no record is the one outcome there is no
  way back from.
- **Idempotent on the session id** (`unique` on the column) — at-least-once
  delivery makes a duplicate the ordinary case rather than the edge.
- **Status transitions now have somewhere to go**, and four Stripe events use
  them — see below.
- **FX stays NULL.** The primary-currency equivalent comes from the balance
  transaction, which is not on this event, and the four FX columns are
  all-or-nothing. A gift in the wedding's own currency never needs them.

**The idempotency key is a hash of the whole request plus a five-minute bucket.**
It used to fold the guest's note down to its LENGTH, which had both failure modes
a key exists to avoid: two different gifts of the same amount from one household
collided, so the second silently never charged; and a retry with different words
hit Stripe's `idempotency_error` — permanently, because the key never changed,
dead-ending that guest for 24 hours behind a message about their connection
(S-H1). Now a double-tap is one attempt and everything else is its own.

**Contributions are kept for one year.** At the end of it the couple get a
parting summary — a final count for their records — and the rows go. The
retention window and the personal data in them are in `wiki/compliance/`.

## The four endings a payment can have

A checkout session is not over when it completes, and the first cut of this
webhook behaved as if it were. Four more events are handled, because three of
them decide whether a gift exists at all (S-M2):

| Event | What it means | What it writes |
|---|---|---|
| `checkout.session.async_payment_succeeded` | A delayed bank debit cleared — BECS here, SEPA in Europe | `succeeded`, same path as `completed` |
| `checkout.session.async_payment_failed` | That debit bounced | `failed` |
| `checkout.session.expired` | The guest opened checkout and closed the tab | `failed` |
| `charge.refunded` | The whole charge went back | `refunded` |

The first is the money bug of the four. A BECS or SEPA session completes in
seconds and settles days later, so without it a **real gift the couple were
really given sat `pending` forever** and never reached their log.

`charge.refunded` carries a charge, not a session, so the refund path finds its
row by payment intent — migration 0059 indexes that column, deliberately not
`UNIQUE`, since it is null on every unsettled row and Stripe's own idempotency
already covers the replay. A **partial** refund is acknowledged and left alone:
a couple who returned half a gift still received the other half, and calling the
whole thing refunded would tell them otherwise.

**Every transition is one-way and guarded in the service rather than the route.**
Only a `pending` row may fail, only a `succeeded` row may refund, and the row is
kept rather than deleted. A replayed or forged `expired` from a hostile
connected account answers `ignored` and cannot un-settle a gift somebody
actually gave.

What the couple see of an unhappy ending: a `failed` gift is **absent** from the
gift log, because money that never moved is not a gift and a guest who abandoned
checkout never meant to tell them anything. A `refunded` one stays **visible and
out of the totals** — a refund is a thing that happened to their record, not an
erasure of it. The retention summary counts neither.

The ownership check behind all three is now **one query instead of two** — a
`LEFT JOIN` on the settings row, inside a handler Stripe gives twenty seconds
(P-I1).

## The panel

Takes no card details and never could — the button hands off to Stripe's own
page — and it says so *before* the guest presses, along with who reads the name
and note they wrote: the couple, never the other guests.

Amounts are typed in **major units** and converted once through the currency's
real exponent: JPY has no minor unit and KWD has three, so a fixed ×100 would be
wrong by 100× on a payment screen. An amount the server would refuse cannot be
submitted at all.

The URL the button navigates to must be **Stripe's own checkout origin** before
`location.assign` sees it. It is a full-page navigation out of a payment flow to
a server-chosen string; one bad response body would otherwise make "Continue to
payment" an open redirect (S-L1).

Coming back, `?gift=thanks` thanks the guest **without claiming the money
moved** — anyone can type that query parameter, and only the webhook knows
(S-L2). `?gift=cancelled` says nothing was charged.

`PublicRegistryDto.cashGiftsEnabled` is now the couple's intent **ANDed** with
Stripe's capability, so the guest surface never reasons about the two
separately.

## Tests

`cire/api` 1814 pass, `cire/invites` 997 pass, both builds green. New: the
contribute route's gates, its bounds, its pending row and that a turned-away
guest never reaches Stripe; the `Stripe-Account` header that makes the charge
direct (the only layer it exists at, and its absence would put gift money in
cire's balance); the webhook's settle path, its idempotency, its account and
session re-checks and its pending-vs-paid split; the intent-AND-capability rule
on the guest payload; the panel's hand-off, its currency conversion, its mount
gate, the credentials flag on its call, both arrival banners and every refusal
message. Then the four endings: each event's happy path, each transition's
guard from every wrong starting status, the partial-refund carve-out, a forged
`expired` on somebody else's account, and that the gift log hides `failed`
while keeping `refunded` out of the total.

Verified in a browser against a stub API: the panel renders for a claimed guest
on a wedding taking money, and the thanks banner renders on return.

The `prep-pr` gate ran on this branch. Eleven findings were fixed here; three
are filed in the tracker.

**Security**

- **S-C1** — the origin guard read a missing `Origin` header as same-origin. A
  cross-site form post does not always send one, so a match is now required on
  every state-changing method, and only ARC-signed internal routes are exempt.
- **S-H1** — the webhook decoded the whole body before checking the signature.
  It is bounded at 64 KiB, and a longer body is a 400 before a byte is parsed.
- **S-M1** — the idempotency key bucketed by wall clock, and a bucket has
  edges: two presses a second apart could land either side of one and open two
  payment pages for one gift. The route now reads the guest's own open attempt
  back out of D1 and hands the same page over; Stripe is asked whether that
  page is still payable, and a spent one is replaced. The clock bucket stays as
  the second belt, for the simultaneous pair where no row exists yet.
- **S-L1** — the gift log recorded what the guest was shown, never what Stripe
  charged. Nothing in this checkout can make the two disagree — one fixed line
  item, no promotion codes, no adjustable quantity — so a disagreement means
  something we do not model. It is logged at settle. The figure is not
  rewritten: the money moved either way, and the as-given amount is the one the
  guest agreed to.

**Performance**

- **P-W1 / P-W2** — the retention pass wrote one summary row per wedding in its
  own statement. Two aggregates and one grouped batch now, and a failure warns
  instead of taking the pass down.
- **P-I1 / P-I2** — the contribution reads carried columns nobody used, and the
  reuse lookup rides an index rather than a scan.
- **P-I3** — the give-money panel is lazy, so a guest who only browses the list
  never downloads it.

**Compliance**

- **C-M1 / C-M2** — the data map gains the contribution rows, and the status
  column that makes them render.

Filed, not fixed: the refund lookup takes the first row on a payment-intent
index that is not unique (one intent per session makes a duplicate
unproducible); the Stripe session is created before the row, and an orphaned
unpaid session expires at Stripe on its own; and the year-end gift summary has
no read path yet, which is a gate on turning `STRIPE_SECRET_KEY` on rather than
on this PR.

## Still to land

The organiser's connect panel and the labelling of `failed` / `refunded` in the
portal were the two items here; both landed on #771, the PR stacked on this one.

- The FX capture. The primary-currency equivalent lives on the balance
  transaction, which is on none of these events.
- Nothing is verified against real Stripe: that needs a platform account, a test
  key and a webhook endpoint. Everything ships inert until those exist.


chav-aniket:

---

PR #771, opened 2026-08-24
feat(cire-host): give the registry its settings (cash gifts, 3/n)

### LINKED ISSUE #699: Registry cash gifts — Stripe Connect (Express): onboarding, hosted Checkout, webhook, FX capture
Nested under **Registry** in `cire/wiki/todo/platform.md`.

Cash gifts are refused today: `PUT /registry/settings` answers 409 `stripe_not_ready` on `cashGiftsEnabled` until Connect can actually take a charge. This is that work — Express onboarding for the couple, hosted Checkout for the guest, the webhook that records a contribution, and capturing the FX rate from the balance transaction so the snapshotted primary-currency amount is a real number rather than an estimate.

`registry_contributions` already carries the amount as given plus a primary-currency equivalent written once and never recomputed, so the schema is waiting for this.

Missed by the migration: the parser reads a checklist item only at column 0, and the parent `Registry` line is marked `- [~]` — neither `[ ]` nor `[x]` — so it was not an item and its nested children were never seen.

chav-aniket:
Stacked on #762 (which is stacked on #760, on #759).

## What

The settings the guest surface has been reading all along, given somewhere to be
made. Four decisions: whether the list is published, what it says at the top,
where parcels go, and whether guests may give money instead.

They are the registry module's **third sub-tab** — List / Gifts / Settings —
deep-linkable at `#/w/<id>/registry/settings`. It reads the same cached snapshot
the list does, so moving between sub-tabs costs no fetch, and a save patches the
settings row in place rather than invalidating the whole thing.

## Issues

**Closes**

| Issue | What lands here |
|---|---|
| xchromo/osn#699 | Registry cash gifts over Stripe Connect. This is the last part of the stack: the organiser's settings surface, the publish gate, the owner-only Connect panel, and the gift log. Onboarding and the webhook landed in 1/n, the guest's give-money path in 2/n |

Closes xchromo/osn#699.

**Opened**

| Issue | Finding |
|---|---|
| xchromo/osn-tracker#502 | S-L2 — a stale tab's settings save reverts another editor's changes, including re-publishing an unpublished list or restoring a cleared shipping address. Filed rather than fixed |

## Publishing is blocked while the list is empty

And the notice says why: guests reach the list — and the money-gift option with
it — only once it is published, so an empty published list is a page that
promises a registry and shows nothing.

An **already-published** list keeps the toggle live. Otherwise deleting the last
gift would freeze a couple on "Published" with no way back.

## Stripe onboarding is owner-only, and an editor sees that

Everything else on the tab is `weddingEditor`, because a co-host helping with
the list may write it. Naming the bank account gifts are paid into is not
ordinary help — that route is owner-only at the API.

So the panel is **visible to an editor, disabled, with the reason**. A missing
section answers nobody's question; a co-host who wonders why money gifts are off
gets a straight answer.

## Intent and capability stay two things

"Let guests give money" is disabled until Stripe can take a charge, and the
API's own `stripe_not_ready` 409 is answered by saying that, rather than the
generic "check the fields".

One live Stripe read happens on mount, and only for a couple **mid-onboarding**
— they have just come back from Stripe and the webhook can be seconds behind
them. Nobody else pays for it.

Stripe now returns them to this tab (`registry/settings`) rather than to the
gift list: the button they pressed lives here, and so does the panel that reads
their new status.

## The gift log stopped printing column values

The two tables behind that log share a `status` column and share none of its
values: a claim is `reserved` / `purchased` / `released`, a contribution is
`pending` / `succeeded` / `refunded`. The view rendered the column raw, so a
couple met the word "succeeded" about a wedding present.

`giftStatus()` gives each one a word they would use — Promised, Bought, No
longer coming; Not cleared yet, Received, Refunded — and tints the two endings
that mean the gift is not coming. A refunded gift also gets a sentence, because
a one-word pill cannot carry both facts: it went back to the guest, and it is
not in the total above. A value this build has no word for is shown as it came
rather than blanking the row's only state.

`failed` has a word for completeness only: the API leaves those rows out of the
log, since money that never moved is not a gift.

Four fixes landed after the first performance and security pass.

**The Connect response decides where a signed-in owner's browser goes**, so it
is parsed first: `isStripeConnectUrl` accepts `https://connect.stripe.com` and
nothing else, the twin of `isStripeCheckoutUrl` on the guest site (S-L1).

**A disabled control hides its own explanation** (C-L2). The Connect button an
editor cannot press is `aria-disabled` rather than `disabled` — a screen reader
in forms mode skips a `disabled` button, which put the reason out of reach of
exactly the person asking for it. It keeps its tab stop and its
`aria-describedby`; `connectStripe()` refuses on `canManage` and the owner-only
route refuses behind that. The publish radio keeps native `disabled`, because
Solid compiles `checked={published()}` to an effect that only re-runs when the
signal changes and a click on an `aria-disabled` radio would desynchronise it
for good — its notice hangs off the Visibility fieldset instead, where it is
announced on entering the group.

**The docs claimed a Stripe probe that does not exist** (C-L1). Nothing in
`@cire/host` reads a capability flag: the Money-gifts panel renders on every
tier and a keyless deployment surfaces itself as a 404 toast. Corrected in
`registry-stripe.ts`, `wiki/compliance/access-control.md` and
`wiki/systems/cire-registry.md`, with what it would take to make the old claim
true.

**Two stale chunk comments in `ModuleShell`** (P-I2): the registry is two chunks
across three sub-tabs now, and the settings sub leans on `warmPanel`'s
hover/focus prefetch rather than sharing the list's chunk.

A second pass over the same branch found five more, all fixed here.

**Two owner-only Stripe calls were reachable by a co-host** (S-L3, S-L4). "Check
again" and the mount-time refresh both hit the owner-only route, so for an editor
they could only ever come back 403 — which the panel rendered as "Couldn't reach
Stripe just now", an outage message for a permission answer. Both now check
`canManage` first, and on mount that gate sits ahead of `claimStripeCheck`, which
spends a one-shot token: checking it before a gate that refuses would burn the
owner's single check of the page load on a request that never happens.

**`aria-disabled` was decoration** (S-L5). The Button styled it and its comment
claimed the handler no-ops, but the component spread props straight onto the
element, so the click fired and every caller was on its own. The click is now
swallowed in the component — `preventDefault` included, so a `type="submit"`
button in a form does not submit — and Solid's bound-tuple handler form is
dispatched as well as the plain function.

Two rendering costs (P-I3, P-I4): `seed()` sets seven signals and both callers
reach it past an `await`, outside Solid's automatic batching, so the whole form
rendered seven times on load and seven more after every save; and `giftStatus`
reads nothing but its argument, so it moves to module scope rather than being
rebuilt on every render of a view that re-renders on every keystroke.

One finding was filed rather than fixed: **S-L2**, a stale tab's settings save
reverting another editor's changes — including re-publishing an unpublished list
or restoring a cleared shipping address —
[osn-tracker#502](https://github.com/xchromo/osn-tracker/issues/502).

## Still blocked on paperwork

`STRIPE_SECRET_KEY` must stay unset on every tier until the Stripe SSA, DPA and
Connect platform terms are accepted — the gate in
`wiki/compliance/subprocessors.md`. Nothing here changes that; with no key the
Connect routes are not mounted at all.

## Tests

`cire/host` 1151 pass (84 files), `cire/api` 1804 pass. New: the sub-tab's route
parsing and deep link, the empty-list publish gate and its already-published
exception, the editor's disabled Stripe panel and its reason, the
`stripe_not_ready` message, the mount-time refresh firing only mid-onboarding,
the settings save patching the cached snapshot, the Connect URL refusal, and
every gift status reading as a word rather than a column value.

https://claude.ai/code/session_01ESA1nkHzSSaTNzYhfxJq3a


chav-aniket:

---

PR #818, opened 2026-08-28
chore(scripts): guard the jest-dom suppression markers in CI

chav-aniket:
## Summary

`perf/jest-dom-setup` (#792) stopped `vite-plugin-solid` from injecting `@testing-library/jest-dom/vitest` into every Vitest project, using the only opt-out the plugin offers: a `setupFiles` entry whose path matches `/jest-dom/`. That suppression is one line per config, and a config that loses it silently goes back to injecting. Nothing fails; the suite just gets slower again. `shared/test-config/no-jest-dom.ts` has the same problem from the other end — it carries no `package.json`, so turbo does not hash it, and real setup code landing there would go stale in thirteen packages' caches at once.

This branch adds the guard.

- `scripts/check-jest-dom-markers.ts` (471 lines) reads every `vitest.config.ts` in the repo **as text** and fails when a config that imports `vite-plugin-solid` declares no `setupFiles` entry matching `/jest-dom/`. It also fails when the marker file grows anything beyond comments and `export {};`.
- Wired into CI as a step in the existing `script-tests` job, and into `package.json` as `check:jest-dom-markers`.
- The convention is written down: a Conventions row in `CLAUDE.md`, a section in `wiki/conventions/testing-patterns.md`.

Text, not import, is deliberate: the `script-tests` CI job runs no `bun install`, so the script imports nothing outside `bun:test` and Bun/Node built-ins.

The invariant it enforces is not "no marker means injection". It is that **every `vitest.config.ts` loading the plugin declares its jest-dom stance explicitly**.

## Workspaces affected

None. The diff is `scripts/`, `.github/`, `CLAUDE.md`, `wiki/`, and one root `package.json` script entry. No package's source or tests change.

### Closes

| Issue | What |
|---|---|
| `xchromo/osn-tracker#504` | A guard that stops a jest-dom-only setup file coming back. |
| `xchromo/osn-tracker#530` | `P-I1` — the marker file sits outside every package's turbo `test` input hash, so nothing but its own comment stops it growing real setup code. |

Closes xchromo/osn-tracker#504.
Closes xchromo/osn-tracker#530.

### Opened

None. Every finding from the three `/prep-pr` reviews was fixed in this diff or explicitly ruled no-change — see **Review findings** below.

### Parse the configs as text, not by importing them

**Issue:** The guard has to know which configs load `vite-plugin-solid` and what their `setupFiles` says.
**Why:** Importing a `vitest.config.ts` needs the whole dependency tree resolvable. The `script-tests` job checks out the repo and runs `bun test scripts/`, with no `bun install` — an import would either fail there or force an install into a job that deliberately has none.
**Solution:** A small bracket-matching reader over the file's text: `mask()` blanks strings, comments and regex literals; `blockOf()` slices a named object; `setupFilesValue()` pulls the value in any of its shapes.
**Rationale:** The guard checks a syntactic property — does this file name a jest-dom path — and a syntactic property can be checked syntactically. It buys CI independence at the cost of not seeing through indirection, which is stated in the code and covered under **Known limit** below.

### Two exemptions, both narrow

**Issue:** Not every config that lacks a marker is wrong.
**Why:** A config that never loads the plugin is never injected into; and the plugin skips a project running `browser: { enabled: true }` itself, because browser mode bundles its own assertions.
**Solution:** Those two cases return no finding. The browser check runs against `topLevel(browser)` — the block with nested `{…}`/`[…]` blanked — so an `enabled: true` buried in `browser.instances[]` cannot forge the exemption.
**Rationale:** A guard that fires on correct configs gets suppressed, and a suppressed guard is worse than none.

### Fails on a marker file that grows code

**Issue:** `shared/test-config/no-jest-dom.ts` works because of its *name*. Nothing stops someone treating it as the setup file it looks like.
**Why:** It belongs to no package, so turbo does not hash it into any `test` input. Setup code there would run against thirteen stale caches.
**Solution:** `checkMarkerFile()` strips comments and accepts only `export {};`.
**Rationale:** The comment in that file already says this. The guard makes the comment enforceable.

### Known limit

A config reaching the plugin through a **local wrapper that re-exports it** is not detected, and cannot be from text alone — the specifier `vite-plugin-solid` never appears in the file. No config in this repo does that; the convention is to name the package directly. This is documented in the `IMPORTS_SOLID_PLUGIN` JSDoc rather than papered over, so whoever writes the first wrapper reads why the regex needs their specifier added.

## Review findings

All three `/prep-pr` reviews ran against this diff. Findings landed on unmerged code, so they were fixed here rather than filed.

| Review | Finding | Disposition |
|---|---|---|
| Security | None. | No new trust boundary in the CI step; `isDirectory()` reads the dirent so a symlinked directory is never followed; `Bun.spawn` takes array argv; no new dependencies. |
| Performance | `P-I1` serial `readdir` walk | Fixed — `Promise.all` fan-out over subdirectories. |
| Performance | `P-I2` six subprocess spawns in the CLI test file | No change. Reviewer agreed; the file's header comment already justifies it — a CLI contract is only testable through the CLI. |
| Tests | Bare `setupFiles` written as a template literal was not matched | Fixed in `setupFilesValue`, plus a test. |
| Tests | An `enabled: true` nested inside `browser.instances[]` forged the browser exemption | Fixed with the new `topLevel()` helper, plus a test. |
| Tests | `import("vite-plugin-solid")` and `require("vite-plugin-solid")` evaded the import check | Fixed by widening `IMPORTS_SOLID_PLUGIN`, plus two tests. |
| Tests | `SKIP_DIRS` members other than `node_modules`, and the `projects` shape, were untested end to end | Fixed — two CLI tests, one putting a marker-less config under `dist/`, one exercising a two-project config with the browser exemption. |
| Tests | A config reaching the plugin through a local re-exporting wrapper | Not fixable from text. Documented in the `IMPORTS_SOLID_PLUGIN` JSDoc — see **Known limit**. |

## Test plan

| Gate | Result |
|---|---|
| `bun run check` | pass |
| `bun run lint` | pass — 0 errors |
| `bun run fmt:check` | pass — 1416 files |
| `bun run test` | pass |
| `bun run test:scripts` | pass — 86 tests across 8 files, 125 assertions |
| `bash scripts/validate-changesets.sh` | pass |
| `bash scripts/changeset-required.sh` | pass |
| `bun run check:jest-dom-markers` | `32 vitest configs checked, jest-dom suppression intact.` |

`scripts/` is not covered by `bun run check`, so it was typechecked separately with a throwaway `tsconfig` extending the root config over `scripts/**/*.ts` — exit 0, config deleted, `git status` clean afterwards.

The guard has 28 unit tests over `mask` / `checkJestDomMarkers` / `checkMarkerFile` and 8 subprocess tests over the CLI itself — exit codes, stderr, the config count, the skip list, and the `projects` shape. Both negative directions are covered: a missing marker is a finding, and each exemption suppresses one.

### Verified it bites

Run against the tree with `perf/jest-dom-setup`'s markers in place: clean, 32 configs. The unit and CLI tests cover the failing direction on fixtures rather than by mutating the repo — a marker-less config produces a finding naming the file and, in the `projects` case, the project.


chav-aniket:
Closing in favour of #827, which carries the two lines from this PR that earn their keep.

The justification here was that `vite-plugin-solid` probes for `@testing-library/jest-dom` with `require.resolve` from its own module context, so a package could drop the dependency from `devDependencies` and stay green on Bun's hoisting — a green suite over a broken dependency graph. That trap does not exist on this install.

Bun 1.4's layout is not hoisted. There is no root `node_modules/@testing-library/`; each package gets its own symlink into `node_modules/.bun/`, and a package that does not declare a dependency cannot resolve it. Probed from a package with the devDep removed:

```
FAIL  Cannot find module "@testing-library/jest-dom/vitest" from "osn/api"
```

and through Vitest itself:

```
Error: Cannot find module '/…/node_modules/@testing-library/jest-dom/vitest'
```

Red, not green. The failure this guard was built to catch announces itself.

What is left is worth keeping, and #827 has it in about forty lines instead of 1022:

- `globalDependencies: ["shared/test-config/no-jest-dom.ts"]` in `turbo.json`, so an edit to a file no package owns cannot be served from cache everywhere.
- `scripts/check-jest-dom-markers.sh`, a grep that fails the build when a Solid Vitest config loses the `setupFiles` marker — which is the way this actually regresses, and the one failure mode that is genuinely silent.

The devDependency prune this PR's premise implied is also in #827: ten packages declared `@testing-library/jest-dom` while importing no matcher, and they are now the three that do.

xchromo/osn-tracker#504 and xchromo/osn-tracker#530 are closed by #827.

PR #821, opened 2026-08-28
fix(cire): mail the couple their gift summary, and fix the two gift-path flaws under it

### LINKED ISSUE #787: Send the couple a year-end summary of their registry gifts
Registry contributions in cire are kept for one year. At the end of that year the retention pass (`writeGiftSummaries` in `cire/api/src/services/retention.ts`) rolls each wedding's gifts into a summary — count, total, currency — and deletes the individual rows, names and messages included.

The couple should get that summary rather than simply lose the detail. Two pieces are missing:

1. **An address to send to.** cire holds no organiser email. The wedding knows its owning OSN account; the address lives in `@osn/api`. That needs a new ARC-authenticated internal route on osn-api — something like `POST /internal/accounts/emails` taking account ids and returning addresses — which cire calls with its existing ARC signer (`@shared/crypto`, see `wiki/systems/arc-tokens.md`).
2. **The mail itself.** A transactional template through `@shared/email`, which is already live on Resend from `hello@cireweddings.com`. Content is the count, the total in the wedding's primary currency, the date range, and a line saying the individual entries have now been erased.

Alternatively, or as well, surface the summary in the host portal next to the gift log so it is reachable without an email at all.

**Done when** a couple whose gift detail has been erased can still see, or has been sent, the count and total of what they were given.

**Not urgent:** `STRIPE_SECRET_KEY` is unset on every tier, so no gift has been taken and no retention clock is running.

chav-aniket:
## Summary

The retention sweep already deleted a wedding's per-gift detail a year after its last event, and left the couple an aggregate they could not see. This PR closes that loop: the sweep now mails the couple a summary at the moment the detail goes, and the organiser's registry settings show the kept record afterwards. Two flaws in the gift path found during review go with it — a refund settled the wrong contribution when two rows shared a payment intent, and a Stripe checkout session could open for a gift that was never written.

Reaching the couple needs an address, and cire stores none of its own, so this also adds one small route to osn-api: `POST /internal/accounts/emails`, ARC-authenticated behind a new `account:email-read` scope, capped at 100 ids per call, restricted to cire-api at the point of use, and silent about ids it cannot answer for so it cannot be used as an existence oracle.

- Mails one `registry-gift-summary` email per swept wedding, best-effort, never blocking the delete.
- Adds a read path for the kept summary in the organiser's registry settings.
- Fixes the refund lookup to settle the contribution that names the payment intent.
- Writes the contribution row before opening the Stripe checkout session.
- Bounds the sweep at 250 weddings per run and its address lookups at 4 calls in flight.

## Workspaces affected

| Package | Change | Changeset |
|---|---|---|
| `@cire/api` | Gift summary delivery, refund and ordering fixes, sweep bounds | `.changeset/gift-summary-delivery-cire.md` (minor) |
| `@cire/db` | `0060_contribution_session_nullable.sql` + schema | same |
| `@cire/host` | Registry settings read path for the kept summary | same |
| `@osn/api` | `POST /internal/accounts/emails`, `account:email-read` scope | `.changeset/gift-summary-delivery-platform.md` (minor) |
| `@shared/email` | `registry-gift-summary` template + metric | same |
| `wiki/` | `compliance/access-control.md`, `data-map.md`, `retention.md` | not a package |

`scripts/validate-changesets.sh` passes: the two changesets are split so no ignored `@cire/*` package shares one with a versioned package.

### Closes

| Issue | What it asked for |
|---|---|
| `xchromo/osn#787` | Send the couple a year-end summary of their registry gifts |
| `xchromo/osn-tracker#527` | S-L2 |
| `xchromo/osn-tracker#528` | S-L3 |
| `xchromo/osn-tracker#529` | C-L1 |

Closes #787
Closes xchromo/osn-tracker#527
Closes xchromo/osn-tracker#528
Closes xchromo/osn-tracker#529

### Opened

Findings from this PR's own review that are not fixed here. All in the private tracker; referenced by number and finding ID only.

| Issue | ID(s) |
|---|---|
| `xchromo/osn-tracker#534` | S-M1 |
| `xchromo/osn-tracker#535` | P-W2 |
| `xchromo/osn-tracker#536` | P-I2 |
| `xchromo/osn-tracker#537` | T-M2 / T-E1 / T-R1 |
| `xchromo/osn-tracker#538` | T-E2 / T-E4 |
| `xchromo/osn-tracker#539` | T-E3 / T-U7 |
| `xchromo/osn-tracker#540` | T-U5 / T-U3 |
| `xchromo/osn-tracker#541` | T-U4 / T-S2 / T-S3 |
| `xchromo/osn-tracker#542` | T-S1 |
| `xchromo/osn-tracker#543` | T-S4 |

Fixed in this PR and so not filed: S-M2 (the issuer pin), S-L1, P-C1, P-W1, P-I1, C-M1, and the T-U1 / T-U2 / T-R2 tests.

### The route takes profile ids, not account ids — `C-H12`

**Issue.** `xchromo/osn#787` describes the lookup in terms of account ids, but cire holds no account id for a wedding's owner. It holds `owner_osn_profile_id`.

**Why.** Making cire translate profile id to account id first would mean either a second internal round trip or cire storing an identifier it has no other use for. The account id is also the identifier osn-api's own invariants keep off every client surface (P6).

**Solution.** `POST /internal/accounts/emails` takes `profile_ids` and resolves each to the address of the account that owns it. The account id never leaves osn-api.

**Rationale.** Fewer identifiers crossing the boundary, and the one that stays behind is the sensitive one.

**Out of scope.** An account-id form of the same lookup, if another service ever wants one.

### An address lookup that never says an id is unknown — `S-M3`

**Issue.** A per-id `not_found` would make the route an existence oracle: anyone holding the scope could probe arbitrary `usr_*` ids and learn which exist.

**Why.** The caller has no use for the distinction. A missing address, a tombstoned account and an id that was never real all mean the same thing to the sweep: no mail.

**Solution.** Always 200, always a plain array of what could be answered. Ids that could not be are simply absent, and the three cases are indistinguishable. The body caps at 100 ids and each id at 64 characters, both enforced by the schema before the handler runs.

**Rationale.** The cap comes from D1's 999 bound-variable ceiling as much as from abuse; the omit-list costs the caller nothing and removes the oracle entirely.

**Out of scope.** Rate limiting the route per caller key — filed as `xchromo/osn-tracker#534`.

### The scope is pinned to cire-api at the point of use — `S-M2`

**Issue.** `PERMITTED_SCOPES` in `osn/api/src/routes/graph-internal.ts` is one flat allowlist governing what `/graph/internal/register-service` will grant to any service. Holding `INTERNAL_SERVICE_SECRET` is therefore enough to register a key asking for `account:email-read`.

**Why.** Making that allowlist per-service is a larger change than this PR should carry, and it would still be worth pinning afterwards: a scope granted by mistake should not be spendable from the wrong service.

**Solution.** `osn/api/src/routes/internal-account.ts` checks `caller.iss === "cire-api"` after the ARC signature verifies, and returns the same 401 body as a missing or wrong-scoped token.

**Rationale.** Enforced where the issuer is signature-verified rather than where it is self-declared. The identical 401 means a service that should not be there learns nothing from the difference.

**Out of scope.** Per-service scope allowlists at registration.

### The summary email is best-effort and never retried — `C-H14`

**Issue.** By the time the email is sent the deletes have committed. What should a failed send do?

**Why.** The retention obligation is discharged by the delete. The email is a courtesy record. A retry loop against a mailbox that is still down would only mail the same couple again on the next sweep with no new data.

**Solution.** `cire/api/src/lib/gift-summary-email.ts` catches per wedding and again around the whole batch, logs a warning, and returns. No address means no mail and no error. Sends run at concurrency 4.

**Rationale.** One couple loses one email; the cohort and the sweep are unaffected.

**Out of scope.** An alert on silent non-delivery. Nothing currently watches for it — see the test plan's negatives.

### Bounds on the sweep and its lookups — `P-C1` / `P-W1`

**Issue.** The sweep processed every eligible wedding in one run and fanned every address lookup out at once.

**Why.** Both are unbounded in the size of the cohort, which is however many weddings passed their year on the same day.

**Solution.** `MAX_WEDDINGS_PER_SWEEP = 250`, ordered by final event so the oldest go first, with a `logInfo` when the run is capped. `MAX_EMAIL_CALLS_IN_FLIGHT = 4` in `cire/api/src/services/osn-bridge.ts` via a windowed chain.

**Rationale.** A capped run is a run that finishes; the remainder is picked up by the next one, and ordering by final event means nothing starves.

**Out of scope.** Making the cap configurable.

## Test plan

| Gate | Result |
|---|---|
| `cire/api` bun test | 1847 pass / 0 fail / 5823 expect() / 109 files |
| `osn/api` vitest run | 67 files / 1086 tests passed |
| `cire/host` vitest | 84 files / 1156 tests passed |
| `cire/invites` vitest | 997 passed |
| `cire/db` bun test | 5 pass / 31 expect() |
| `shared/email` vitest | 59 passed |
| `cire/host` astro check | 194 files, 0 errors, 0 warnings, 2 hints |
| `osn/api` tsc --noEmit | clean |
| `shared/email` tsc --noEmit | clean |
| `bun run fmt:check` | all matched files correct (1436 files) |
| root `oxlint` | only pre-existing warnings, none in files this PR touches |
| `scripts/validate-changesets.sh` | all changeset package references valid |

Honest negatives, in the order they would bite:

Nothing is signed with Stripe yet, so `STRIPE_SECRET_KEY` must stay unset on every tier. The Stripe routes are key-optional and do not mount without it, so merging is safe; setting the key is not, and `/privacy` has to name Stripe as a recipient before it goes on.

The 30-second `Effect.timeout` around the sweep is still untested — filed as `xchromo/osn-tracker#538`. So is the `catch { set.status = 500 }` path on the new osn-api route, and a multi-id request where only some ids resolve, both in `xchromo/osn-tracker#537`.

Address disclosure writes no audit-log entry, because osn-api has no audit table to write one to. It is recorded in `wiki/compliance/data-map.md` and `wiki/compliance/access-control.md` instead. There is also no new metric counter for the lookup, so a silent failure to deliver summaries would not show up anywhere.

Writing the contribution row before the checkout session opens means a fast double-tap can now write two rows where it wrote one; the loser is closed `failed` with a null session. Any "gifts attempted" count an organiser sees will therefore include rows that were never real attempts.

Three `no-await-in-loop` disables remain in `cire/api/src/services/link-preview.ts` (lines 699, 708 and 874). All three are pre-existing on `main` and untouched here. They cannot honestly become `Promise.all`: each redirect hop depends on the previous response's `Location`, and the image loop would resolve candidates it never reaches, defeating a deliberate cap on DoH lookups. Raised separately rather than silenced.


chav-aniket:

---

PR #831, opened 2026-08-28
feat(cire): let the couple download their gift log as CSV

chav-aniket:
## Summary

The couple can now download their whole gift log as a spreadsheet.

`GET /api/organiser/weddings/:weddingId/gifts.csv` returns every claim and every
cleared cash gift for one wedding as a CSV, and a quiet "Download gifts (CSV)"
button in the Gifts tab of the registry module fetches it. It is the third
organiser export, beside `guests.csv` and `events.csv`, and it sits on the same
export route group: `osnAuth()` → `weddingMember` → per-user rate limiter.

The export exists because two things already in the registry are in tension. The
portal reads the gift log **a page at a time** (50 rows, capped offset), and the
retention sweep **deletes the detail after a year**, leaving only the aggregate
summary. Between those two, the couple have no way to hold the whole thing. A
download is that way, and it is what a data-portability request is answered with.

## Workspaces affected

| Workspace | What changed |
|---|---|
| `@cire/api` | New `GET …/gifts.csv` route, `giftExportService`, and a server-side `minorToDecimal` money helper |
| `@cire/host` | The export button in `RegistryView`, and the `weddingSlug` prop that names the file |
| `wiki/` | `wiki/systems/cire-registry.md` documents the route and its ceiling; `wiki/compliance/retention.md` records the downloaded copy our retention sweep cannot reach |

## Issues

Opened during review of this branch (private tracker):

| Finding | Issue | What it is |
|---|---|---|
| S-H1 | `xchromo/osn-tracker#546` | The export is deliberately not entitlement-gated; recorded as an accepted risk with a test to keep |
| S-L1 | `xchromo/osn-tracker#547` | A negative amount would leave the file as text, not a number (latent — no path writes one today) |
| P-W1 | `xchromo/osn-tracker#548` | The two reads apply the ceiling separately, so twice the printable rows cross the wire |
| P-W2 | `xchromo/osn-tracker#549` | The file is buffered whole instead of streamed; the precondition for raising the ceiling |
| P-W3 | `xchromo/osn-tracker#550` | The heaviest export shares the default 10/min export limiter |
| P-I1 | `xchromo/osn-tracker#551` | The handler re-reads a wedding row the auth gate already read |
| P-I2 | `xchromo/osn-tracker#552` | The truncation warning logs a row count that is itself clamped |
| P-I3 | `xchromo/osn-tracker#553` | No check constraint on the contribution currency columns |
| T-S3 | `xchromo/osn-tracker#554` | Parity with the portal's gift log is asserted only in prose |
| T-S4 | `xchromo/osn-tracker#555` | The ceiling test checks the row count but not which rows survive |
| T-S6 | `xchromo/osn-tracker#556` | Two `money.ts` branches are unreached by its tests |

Fixed in this branch rather than filed: S-M1 (`EXPORT-D4`), P-C1 (`EXPORT-D2`),
C-L1 (the retention-register row), and the six test findings T-S1, T-S2, T-S5,
T-S7, T-S8 and T-R1.

### The file matches the screen, exactly — `EXPORT-D1`

**Issue.** The export could reasonably have differed from the portal: drop the
host family, drop refunded gifts, add the Stripe references.

**Why.** Every difference is a second disclosure surface that nobody reviewed. A
gift log the couple read on screen has been through a threat model; a file that
quietly holds more, or less, has not.

**Solution.** `giftExportService.giftsCsv` reads the same two tables in the same
shape as `registryService.giftLog`: the same `failed`-contributions exclusion
(money that never moved is not a gift, while a `refunded` gift did happen and
stays visible), the same LEFT join for a cash gift with no item, and no
host-family exclusion. No Stripe identifiers — the charge and payment-intent
references are payment plumbing, not part of the couple's record of who gave what.

**Rationale.** One reviewed shape, two renderings of it. The one place the
promise is not yet enforced by a test is filed as T-S3.

### A ceiling that says so, not silent truncation — `EXPORT-D2`

**Issue.** The gift log is unbounded in principle: every household may claim
every item and send cash on top. An export cannot page — the whole point of the
download is that it is the whole record.

**Why.** An unbounded read of a four-figure log is exactly the shape that turns a
free-tier D1 read budget into an outage, and it is also the shape a rate limiter
cannot save you from at one request.

**Solution.** `MAX_GIFT_EXPORT_ROWS = 2000`, and a read over the ceiling emits
`Effect.logWarning` annotated with the wedding, the row count and the cap rather
than truncating in silence. The number comes from the **CPU budget, not from how
many gifts a wedding plausibly has** — the first draft used 5,000 on that
reasoning and the performance review (P-C1) showed it does not fit. Workers Free
allows 10 ms of CPU per invocation, and building the file is dominated by
`serialiseCsv`: roughly 6 ms for 2,000 rows and 11 ms for 5,000, so the old
ceiling spent the whole budget and the request died with `exceededCpu`.

**Rationale.** A portability answer that quietly drops rows is worse than no
answer at all, and a ceiling the runtime cannot reach is not a ceiling. Raising
it back toward "a plausible wedding" means streaming the response first — that is
P-W2, filed against this file.

### Amounts are numbers, and the exponent is not 2 — `EXPORT-D3`

**Issue.** A CSV of money is opened in a spreadsheet and summed.

**Why.** `$125.00` is text to a spreadsheet. And the minor-unit exponent is not
always 2: JPY has no minor unit at all, KWD/BHD/JOD use three, so a fixed `/ 100`
mis-states an amount by 100× the moment a gift arrives in yen.

**Solution.** Bare major-unit decimals with the currency in its own column, and a
new `cire/api/src/lib/money.ts` that reads the exponent off
`Intl.NumberFormat(...).resolvedOptions().maximumFractionDigits`, memoised —
including a remembered rejection, so an unknown code is tried once and falls back
to 2 rather than throwing per row. It mirrors the host-side helper of the same
name; the API had none.

**Rationale.** The FX columns carry the equivalent snapshotted at charge time and
are never recomputed: a gift log that re-values itself as rates move is not a
record of anything.

### A household is named, never coded — `EXPORT-D4`

**Issue.** The first draft printed `families.public_id` beside the household
name, on the reasoning that a stable identifier helps the couple match rows.

**Why.** That column is the claim code. It is a bearer credential: on its own it
opens that household's invite, and the `HOST-*` code opens every event. Printing
it turns a thank-you list into a file of working sign-ins sitting on a laptop or
in a mail attachment — and `wiki/compliance/data-map.md` already classifies it as
a Credential (S-M1).

**Solution.** The column is gone. The file carries the household name and the
giver's chosen display name, which is what a thank-you list actually needs, and
which is exactly what the portal's gift log returns. Two route assertions pin it:
the header does not contain `Household Code`, and the body does not contain a
seeded claim code.

**Rationale.** The export must not disclose more than the screen it copies
(`EXPORT-D1`), and a credential is the one field where "more" is not a matter of
degree.

### The export is not gated on the registry entitlement — `EXPORT-D5`

**Issue.** Every other registry route calls `weddingEntitlement(db, "registry")`.
This one does not, which the security review raised as S-H1.

**Why.** A couple whose entitlement has lapsed must still be able to take their
own gift record out. An export that fails closed turns a billing state into a
data-portability refusal, which is the wrong answer to give about data we are
still holding. The exposure is also small: a wedding that never had the
entitlement has no gift rows, so the export returns the header line and nothing
else.

**Solution.** Left ungated, deliberately, and recorded as an accepted risk in
`xchromo/osn-tracker#546` — which also asks for the comment at the route and the
test that stops a later blanket gate from closing it by accident.

**Rationale.** The read is bounded, rate-limited and authenticated as a member of
that wedding. Refusing to hand back a record we still hold is the worse failure.

### Out of scope

- **No date range, no column picker.** The whole log, or nothing. Filters are a
  feature request, and every one of them is a new shape to review against the
  portal (`EXPORT-D1`).
- **No scheduled or emailed export.** The retention sweep already mails the
  parting aggregate; this is the couple pulling the detail on demand.
- **Nothing about the retention window itself.** Contributions keep the one-year
  guest window; see `wiki/compliance/retention.md`.
- **No streaming.** It is the right shape for this endpoint and it is what would
  let the ceiling follow the data, but it changes `csvAttachment`, which all
  three exports share. Filed as P-W2.

## Test plan

| Gate | Result |
|---|---|
| `cire/api` — `bun test` | 1856 pass, 0 fail (110 files) |
| `cire/host` — `bun run test` | 1162 pass, 0 fail (84 files) |
| `cire/host` — `bun run check` (astro check) | 0 errors, 0 warnings, 2 pre-existing hints |
| `oxlint` on the changed files | Clean (one pre-existing `no-map-spread` warning at `organiser-weddings.ts:110`, outside the diff) |
| `bun run fmt` | Clean, 1444 files |
| `scripts/validate-changesets.sh` | Pass |
| `scripts/verify-vendored-anti-slop.sh` | Pass |
| `bun run build --filter=@cire/api --filter=@cire/host` | Both pass |

New tests: seven route tests (401 / 403 / 404 / 200 with headers, row order and
exclusions / header-only / co-host allowed / viewer co-host allowed, plus the two
assertions that no claim code reaches the file), nine service tests (empty log,
merge order and bare decimals, `failed` dropped and `refunded` kept, JPY exponent
and the FX columns, wedding scoping, formula escaping on guest-written text, the
host household kept where the guest export excludes it, thanked-at apart from
received-at, and the ceiling warning at cap + 1), five money tests, and seven host
tests (no button with nothing to export, the download under a slug-named file,
the failure toast, both auth exits, the in-flight guard, and `weddingSlug`
forwarding through `ModuleShell`).

**Honest negatives.** `cire/api` has no `check` script, so this branch has no
API-side typecheck gate — the same gap every cire API PR has. Nothing here was
run against a real Stripe account: `STRIPE_SECRET_KEY` is unset on every tier, so
`registry_contributions` is empty in production and the cash-gift half of the
export was exercised only against seeded rows. The button was not clicked in a
browser; it is covered by the host tests and by nothing else. The 6 ms / 11 ms
figures behind the row ceiling come from the performance review's measurement of
`serialiseCsv`, not from a profile of the deployed Worker.


chav-aniket:

---

PR #835, opened 2026-08-30
chore(guards): close the gaps around the dependency and vendored-plugin checks

chav-aniket:
## Summary

The repository has a set of guards around its dependencies and its one vendored dependency — a soak policy, a checksum manifest, a changeset gate, a CI workflow. This branch does not change any of the rules those guards enforce. It closes the gaps around them: the paths a change could take to route past a guard without a review, the guard that only ran in CI, the one that certified the vendored tree against itself, and the compliance record that described all of it more confidently than the code warranted.

- New guard `scripts/verify-vendored-anti-slop-pin.sh`, wired into CI on pull requests, with 11 of its own tests.
- `.github/CODEOWNERS` rewritten to cover the dependency and deployment surface.
- `.github/workflows/ci.yml` given a `permissions:` block; it had none.
- `check:release-age-excludes` moved into `lefthook.yml` on pre-push, and taught to refuse a second `bunfig.toml`.
- Two compliance pages corrected against what the branch-protection ruleset actually says.

## Workspaces affected

None — this is CI/infra only. No workspace package changes behaviour. `.changeset/dep-guard-coverage-gaps.md` names `@tools/oxlint-house` at `patch`: `lefthook.yml` and `tools/oxlint/anti-slop/*` are off the changeset allowlist so one is required, and that plugin is the package that lives beside the vendored tree these guards cover. Same sole target as PR #832.

## Issues

**Closes**

| Issue | What |
|---|---|
| xchromo/osn-tracker#495 | `S-M1` |
| xchromo/osn-tracker#496 | `S-M2` |
| xchromo/osn-tracker#498 | `S-L1` – `S-L5` |
| xchromo/osn-tracker#499 | `C-M1`, `C-L1` |

Closes xchromo/osn-tracker#495.
Closes xchromo/osn-tracker#496.
Closes xchromo/osn-tracker#498.
Closes xchromo/osn-tracker#499.

**Opened**

| Issue | What |
|---|---|
| xchromo/osn-tracker#562 | `S-L6` |

### Prove the vendored pin moved, rather than fetching upstream — `S-M1`

- **Issue** — `tools/oxlint/anti-slop/SHA256SUMS` proves the tree matches its own manifest. That is a self-certification: edit a vendored file, regenerate the manifest, and the manifest certifies whatever the tree now holds. Nothing checked that the tree and the upstream commit it claims to be a copy of moved together.
- **Why** — it is the one dependency with no registry behind it: no resolved version, no integrity hash from a registry, no advisory feed. The whole integrity claim rests on "this is a verbatim copy of commit `<sha>`", and until now nothing tested the second half of that sentence.
- **Solution** — `scripts/verify-vendored-anti-slop-pin.sh <base-ref>` fails a change to `tools/oxlint/anti-slop` that does not also move the pinned commit in that directory's `README.md`. It runs in the `script-tests` job, gated on `github.event_name == 'pull_request'`, after a `--depth=1` fetch of `github.event.pull_request.base.sha`.
- **Rationale** — the rejected option was fetching the upstream repository in CI and diffing against it. That is the stronger check and it is also a network hop on every PR, a flake source, and a new outbound dependency in a workflow that currently has none. The pin check needs no network and is deterministic. It buys less: it cannot say the new bytes are upstream's, only that a human was asked to look. That is what `.github/CODEOWNERS` on the same path is for. Two accepted costs: a re-vendor to the *same* upstream commit is refused, which is the review the guard wants; and `README.md` and `SHA256SUMS` are exempt from the guard, because the README carries the pin and its own hash lives in `SHA256SUMS`, so editing either forces the other and an unexempted guard would fire on this very PR.

### A separate script, not a branch inside the existing one — `approach`

- **Issue** — the obvious home for the pin check is `scripts/verify-vendored-anti-slop.sh`, which already knows the vendored path.
- **Why** — that script runs in `lefthook.yml` on pre-commit as well as in CI.
- **Solution** — a second script, called only from CI, on pull requests.
- **Rationale** — a pre-commit hook has no base ref to diff against. Folding the check in would mean a conditional that silently does nothing in the local half, which is the shape a guard rots into. Separate scripts make "this one needs a base" a fact of its signature. It exits `2` for a missing or unfetched base ref and `1` for a vendored change with an unmoved pin, so a workflow-wiring mistake reads differently from a vendoring one.

### Base SHA, not merge base — `approach`

- **Issue** — the pin guard diffs two trees, and `actions/checkout` clones at depth 1, so neither the base commit nor a merge base is in the checkout.
- **Why** — computing a merge base means deepening the clone until the histories meet, which is unbounded work on every PR.
- **Solution** — `git fetch --no-tags --depth=1 origin "$BASE_SHA"`, then diff `$BASE_SHA` against `HEAD`.
- **Rationale** — `git diff A B` compares two trees and needs no common ancestor between them, so one shallow fetch of the single base commit is sufficient. The difference from a true merge-base diff only shows on a branch that is behind, where the guard would see the base's own vendored changes as this branch's — and those already moved the pin when they landed, so the guard stays quiet.

### CODEOWNERS covers what a guard runs *through*, not only what it checks — `S-M2`

- **Issue** — the file owned `/.github/` and `/scripts/`. It did not own `bun.lock`, `.bun-version`, any of the 37 workspace `package.json` files, `bunfig.toml`, `oxlintrc.json`, `lefthook.yml`, or the five `wrangler.toml`/`wrangler.jsonc` files.
- **Why** — owning the workflow and not the indirection leaves the cheapest bypass unreviewed. Every CI step calls a guard by script name out of a `package.json`; `oxlintrc.json` decides the vendored plugin is the one that runs; `bun.lock` and `.bun-version` decide which bytes the soak policy actually resolves to; the wrangler files name production routes and bindings, so an edit there can point a live hostname at a different database without touching application code.
- **Solution** — patterns added for all of the above. `package.json` and `bunfig.toml` are deliberately left unanchored so they match at any depth; the rest are anchored with a leading `/`.
- **Rationale** — unanchored is the point for those two. A workspace `package.json` can move a guard behind a different script name, and Bun reads `bunfig.toml` relative to the working directory — `deploy.yml` and several package scripts run with `--cwd`. The header comment records that GitHub applies the *last* matching pattern, which matters where two overlap (`tools/oxlint/house/package.json` matches both `package.json` and `/tools/oxlint/**`); every pattern names the same owner today, so order is not load-bearing yet.

### `permissions: contents: read` on `ci.yml` — `S-L4`

- **Issue** — the workflow declared no `permissions:` block, so it ran with the default token, which is write-all on a repository that has not been switched over.
- **Why** — every job in it reads the checkout and reports through the check run the workflow already owns. Nothing writes an issue, a comment, a release or a package. A compromised dependency running inside a job inherits whatever the token can do.
- **Solution** — a workflow-level `permissions: contents: read`.
- **Rationale** — declared at workflow level rather than per job, because all six jobs want the same thing and a per-job block is a thing to forget on the seventh. The deploy, release and secret workflows already declare their own; this one was the gap.

### The release-age guard runs locally and refuses a second `bunfig.toml` — `S-L3`, `S-L5`

- **Issue** — `check:release-age-excludes` ran in CI only, and it read exactly one file: the root `bunfig.toml`, via `new URL("../bunfig.toml", import.meta.url)`.
- **Why** — CI-only meant an expired `# DROP AFTER` marker was found after a push rather than before one. And Bun does not read the root file; it reads the `bunfig.toml` nearest the working directory. A second one further down the tree would set the install policy for every command that runs with `--cwd` while this guard went on passing against the root file.
- **Solution** — a `release-age` command in `lefthook.yml` under `pre-push`, and a new exported `findStrayBunfigs` in `scripts/check-release-age-excludes.ts` that fails on any tracked `bunfig.toml` below the root.
- **Rationale** — the stray check reads `git ls-files`, not a filesystem glob: the tracked set is what a reviewer sees and what CI checks out, and it does not walk `node_modules`, where a dependency's own `bunfig.toml` is nobody's business. Globbing for every copy and merging their policies was rejected — the repo has never wanted a second file, and a change that adds one is a decision that belongs in a reviewed PR. Outside a git checkout the CLI prints that it skipped the check rather than passing quietly. Six new tests cover it, three unit and three against the real CLI in a throwaway git repo.

### Part of `S-L1` and all of `S-L2` were already fixed — `S-L1`, `S-L2`

- **Issue** — the issue describes an inline `find -type f` pipeline in `ci.yml` with no `set -euo pipefail`, blind to symlinks and to file modes.
- **Why** — worth stating rather than silently closing: a reviewer reading the issue will not find the code it describes.
- **Solution** — that pipeline was extracted into `scripts/verify-vendored-anti-slop.sh` in an earlier PR. It sets `set -euo pipefail` and lists the tree with `git ls-files`, which lists symlinks. So `S-L2` and the symlink half of `S-L1` were closed before this branch. What was genuinely still open is the *generator*: the re-vendor recipe in `tools/oxlint/anti-slop/README.md` still used `find . -type f`, so generator and checker disagreed and a vendored symlink would read as a spurious diff. That recipe now uses `find . ! -type d`, and `SHA256SUMS` is regenerated.
- **Rationale** — the two halves have to agree on what counts as a file or the disagreement surfaces as a mystery diff on the next re-vendor, which is exactly when nobody wants one.

### The compliance pages now say CODEOWNERS routes reviews, not that it gates merges — `C-M1`, `C-L1`

- **Issue** — `wiki/compliance/access-control.md` (item 4) and `wiki/compliance/soc2.md` (item 8, `C-L4`) needed updating for the CODEOWNERS work, and `soc2.md` had no record of the vendored dependency beside its dependency-scanning item.
- **Why** — a compliance page that claims a control is in place when it is not is worse than one that admits the gap.
- **Solution** — both items rewritten against what `gh api repos/xchromo/osn/rulesets/13921406` actually returns, and both `last-reviewed` dates refreshed. Ruleset `13921406` is active and requires linear history, review-thread resolution, and the `CI` and `Require changeset` checks — and it sets `required_approving_review_count: 0` with `require_code_owner_review: false`. So CODEOWNERS routes a review request and does not gate a merge. Required hardware-key MFA and required signed commits are org settings and are unset. A vendored-dependencies block was added under `soc2.md` item 5 (`C-M7`) naming `dmmulroy/anti-slop`, its MIT licence, the pinned commit and the three in-repo checks, ending on what none of them do: compare the tree against upstream.
- **Rationale** — the zero-review setting is deliberate and was decided earlier: there is one maintainer, so requiring an approval would block every merge. The honest record is that the control exists and is not yet enforced, with the thing that unblocks it named. `S-M3` (xchromo/osn-tracker#497) tracks the underlying problem — sole owner, sole author, always-on bypass — and needs a decision, so it is deliberately untouched here.

**Out of scope**

- Neither vendored check reads a file's mode, so flipping `100644` to `100755` passes both. Documented in `tools/oxlint/anti-slop/README.md` and now tracked as xchromo/osn-tracker#562.
- The branch-protection settings that would turn CODEOWNERS into an enforced gate. xchromo/osn-tracker#497.

## Test plan

| Gate | Result |
|---|---|
| `bash scripts/verify-vendored-anti-slop.sh` | ✅ 21 files `OK`, file-set diff empty |
| `bash scripts/verify-vendored-anti-slop.test.sh` | ✅ passed: 7, failed: 0 |
| `bash scripts/verify-vendored-anti-slop-pin.test.sh` | ✅ passed: 11, failed: 0 |
| `bun run check:release-age-excludes` | ✅ passes the release-age guard |
| `bun run test:scripts` | ✅ 61 pass, 0 fail, 89 expect() calls, 7 files |
| `bun run fmt:check` | ✅ 1426 files |
| `bunx --bun oxlint@1.78.0 -c oxlintrc.json scripts/` | ✅ exit 0, warnings only (`no-console`, `no-await-in-loop`), all pre-existing |
| `bash scripts/validate-changesets.sh` | ✅ |
| `bash scripts/changeset-required.sh` (fed the staged file list on stdin) | ✅ `required`, and one is present |
| wikilink resolution on the two touched `wiki/` pages | ✅ no unresolved targets |
| lefthook pre-commit | ✅ lint, format |

Exercised by hand: the pin guard's eleven cases build throwaway git repositories with a base and a head commit, so they cover the vendored-change-with-unmoved-pin failure, the exemptions for `README.md` and `SHA256SUMS`, a missing pin line, an unfetched base ref (exit 2) and a missing base-ref argument (exit 2). The three CLI-level stray-`bunfig.toml` tests likewise build a real git checkout, because the bare temp directories the older CLI tests use make the check report itself skipped.

**Not run:** `bun run lint` and `bun run typecheck` at the repository root. This worktree has no `node_modules` and both root scripts invoke a bare binary, so they exit 127 here; `bun install` was deliberately not run because it prunes `bun.lock`. Lint was run instead through `bunx --bun oxlint@1.78.0` against the changed paths, as recorded above. Nothing in the diff is TypeScript that ships — the two changed `.ts` files are script and test files covered by `bun run test:scripts`, which type-errors would have surfaced at run time.

**Not run:** the new `Check the vendored anti-slop pin` CI step itself. It is gated on `github.event_name == 'pull_request'` and reads `github.event.pull_request.base.sha`, neither of which exists locally; the script beneath it is covered by its own test file, but the workflow wiring — the fetch, the env var, the `if` — is first exercised by this PR's own CI run. The `ci.yml` edits were also not validated by a YAML parser: PyYAML is not installed on this machine, so they were checked by eye and CI is the first parse.


chav-aniket:
@claude this replaces the existing anti-slop guard right instead of adding a new one?

chav-aniket:
> **Claude finished @chav-aniket's task in 29s** —— [View job](https://github.com/xchromo/osn/actions/runs/33306366858)
> 
> ### Checking the guard setup
> * [x]  Read trigger comment and PR description
> * [x]  Verify against actual diff whether the new script replaces or adds to the existing anti-slop guard
> * [x]  Post answer
> 
> No — it's additive, not a replacement. The existing `scripts/verify-vendored-anti-slop.sh` (the checksum check) is untouched and keeps running in both places it always did:
> 
> * `lefthook.yml` pre-commit (`bash scripts/verify-vendored-anti-slop.sh && bunx --bun oxlint ...`)
> * `.github/workflows/ci.yml`, in both the `lint` job (line 35) and the `script-tests` job (line 217)
> 
> The new `scripts/verify-vendored-anti-slop-pin.sh` is a second, separate script, wired into CI only — `.github/workflows/ci.yml` lines 238–244, in the `script-tests` job, gated on `github.event_name == 'pull_request'`. It doesn't run from `lefthook.yml` at all.
> 
> That split is deliberate per the PR's "approach" section: the pin check needs a base ref to diff against, which a pre-commit hook doesn't have. Folding it into the existing script would mean a conditional that silently no-ops locally — "the shape a guard rots into," per the PR body. So the two checks now run side by side and check different things:
> 
> Script	Checks	Runs where
> `verify-vendored-anti-slop.sh`	Tree matches its own `SHA256SUMS` manifest	lefthook pre-commit + CI (`lint` and `script-tests` jobs)
> `verify-vendored-anti-slop-pin.sh`	A change to `tools/oxlint/anti-slop` also moves the pinned commit in `README.md`	CI only (`script-tests`, PRs only)

@claude let's only have guard for the anti-slop guard, consolidate it

chav-aniket:
Consolidation is done and CI is green again.

The bot merged the two scripts but couldn't push `.github/workflows/ci.yml` — GitHub Apps can't modify workflow files — so `script-tests` was still calling the deleted `verify-vendored-anti-slop-pin.sh` and failing with `No such file or directory`. Applied that half in `9ad77b79`.

`script-tests` now runs three steps instead of four:

| When | Runs |
|---|---|
| not a pull request | `verify-vendored-anti-slop.sh` — checksum half only |
| pull request | `verify-vendored-anti-slop.sh "$BASE_SHA"` after a `--depth=1` fetch of the base — both halves |
| always | `verify-vendored-anti-slop.test.sh` |

The `lint` job's call is unchanged: it passes no base ref, which is exactly the checksum-only mode it always ran.

Verified locally rather than trusting the shape: both modes pass against this checkout, and the merged test file is 17 cases, 0 failures — including the four pin cases and `an unfetched base ref exits 2`. All nine checks pass.

One thing I put back: `wiki/compliance/soc2.md` item 5 had lost the word "three" from "three checks that each cover a different failure". Merging two scripts didn't merge two checks — the tree's contents, the pin, and a named owner are still three separate failures being covered.

Issue #598, opened 2026-08-15
Zap: send images, video and voice notes with end-to-end encrypted media storage


chav-aniket:
## What and why

Zap messages can only carry text today. A `c2c` message stores a base64 `ciphertext` of at most about 256 KB (`zap/api/src/lib/limits.ts:12`, `zap/db/src/schema/messages.ts:34-35`), which is too small for video and wasteful for images, and D1 caps a row at 2 MB and a database at 500 MB (`wiki/shared/free-tier-limits.md:188-191`). Media needs its own store.

Because `c2c` chats are end-to-end encrypted, that store must hold ciphertext only: the sender encrypts each file under a fresh key, uploads the encrypted blob, and sends the key plus a pointer to the blob inside the normal encrypted message. Neither the server nor the storage host ever sees the file or the key.

The deferred-decisions table times this "When Zap M2 lands" (`wiki/decisions/deferred-decisions.md:28`), and the Zap page lists the delivery options as direct upload, CDN or peer-to-peer (`wiki/zap/zap.md:55`).

## Proposal

Use Cloudflare R2 for the encrypted blobs, since the stack is already on Cloudflare and `cire-api` already binds R2 buckets (`cire/api/wrangler.toml:207-217`). `@zap/api` issues a short-lived signed upload URL to a chat member, and downloads are gated by chat membership. Blobs for disappearing messages expire with the message. Peer-to-peer delivery is not worth its cost for a first version.

## Done when

- A member can send an image, a video and a voice note in a `c2c` chat, and another member can open it.
- R2 holds only encrypted blobs; a test checks that an uploaded object cannot be read without the key from the message.
- Size caps per media type are set in `zap/api/src/lib/limits.ts`.
- Blobs are removed when their message expires or the chat is deleted.

## Files involved

- `zap/api/src/lib/limits.ts` — size caps
- `zap/db/src/schema/messages.ts` — message storage
- `zap/api/wrangler.toml` — new R2 binding
- `wiki/zap/zap.md` — replace the "Media handling" open question with the design

Issue #597, opened 2026-08-15
Zap: plan message storage past the D1 database limits


chav-aniket:
## What and why

The original question was when to move Zap's message store off SQLite to Postgres or Supabase. That premise is out of date: `@zap/api` runs as a Cloudflare Worker on **Cloudflare D1** (SQLite at the edge). The production database `zap-db-prod` exists (`zap/api/wrangler.toml:67-70`) and the deploy job is live (`wiki/zap/zap.md`, section "CI pipeline").

The real limits are D1's on the Free plan (`wiki/shared/free-tier-limits.md:186-188`, from Cloudflare's published limits — re-check before relying on them):

- **500 MB** per database
- **5 GB** storage for the whole account, shared with the six other databases on it, including `osn-db-prod`, `cire-db` and the dev databases
- **100,000 rows written per day**, account-wide — every message is at least one row write

Messages are the one table that grows without bound. A `c2c` ciphertext can be up to about 256 KB (`zap/api/src/lib/limits.ts:12`), and `messages` has no expiry except the optional `expires_at` for disappearing messages (`zap/db/src/schema/messages.ts:40`). Zap has no users yet, so nothing is at risk today.

## Proposal

Set the trigger before the limit is hit, not after:

1. Add `zap-db-prod` size and daily writes to the free-tier page, with a threshold (for example 50% of either ceiling) that starts the move.
2. When the threshold is reached, pick between: the D1 paid plan (larger databases, higher write quota), sharding messages across several D1 databases by chat, or moving message bodies to object storage (R2) and keeping only metadata in D1. Postgres or Supabase stays an option but means leaving the Workers-plus-D1 stack every other service uses.

## Done when

- `wiki/shared/free-tier-limits.md` names the `zap-db-prod` size and write thresholds and how to check them.
- The chosen path past them is written on `wiki/zap/zap.md`, replacing the "Storage backend" row of its open-questions table.

## Files involved

- `zap/api/wrangler.toml:67-70` — production D1 binding
- `zap/db/src/schema/messages.ts` — the growing table
- `zap/api/src/lib/limits.ts:12` — per-message ciphertext cap
- `wiki/shared/free-tier-limits.md` — D1 ceilings

PR #858, opened 2026-08-31
perf(cire-api): route each request's D1 queries through one D1 session

chav-aniket:
## Summary

Routes every D1 query a request makes through a single D1 session, so that once
read replication is enabled on `cire-db` the second and later queries of a
request can be served by a replica near the reader instead of by the primary in
Oceania.

Today each `prepare()` is an independent query and D1 sends all of them to the
primary. A request that reads three tables pays the Sydney round trip three
times. With a session, the first query pins the bookmark and every later query in
that session can be served by any replica caught up to it — N round trips become
one, plus N-1 local reads.

Measured, from a Sydney client against the Oceania primary: **35.7 ms per
query** (control 15.0 ms, one-select 50.7 ms; p10 33.0, p90 39.3, n=25). From
Europe or North America that figure is several times larger, and it is paid per
query.

**This ships inert.** With no replicas, a session behaves exactly as today —
every query goes to the primary. Enabling replication on the database is a
separate, reversible dashboard change, made after this merges.

### How it threads through

The Elysia app graph is built once per isolate with the Drizzle handle baked in,
and ~50 route factories close over that handle, so a per-request Drizzle client
is not something the app can be handed. Instead the handle is built over a
*stable* client shim (`createSessionRoutedClient`) whose `prepare`/`batch`
delegate to whichever session is current on an `AsyncLocalStorage`, established
per request in `index.ts` by `runInD1Session`. `node:async_hooks` is available
because the Worker already sets `nodejs_compat`.

Outside a session — unit tests, the bun:sqlite dev server, a handler that
somehow escaped the context — the shim falls through to the raw binding, which is
exactly today's behaviour. Losing the context costs latency, never correctness.

Two decisions worth stating:

- **The constraint is `first-primary`, never `first-unconstrained`.**
  `first-primary` sends the first query of a session to the primary, so a request
  always observes every write committed before it started — ordinary
  read-your-writes, including across requests. `first-unconstrained` trades that
  for one saved round trip, and `routes/invite.ts` explicitly cannot take the
  trade: it serves a `no-store`, edit-sensitive payload precisely so organiser
  edits surface on the guest invite's revalidation. One constraint for the whole
  Worker keeps that property from depending on which route was reached.
- **One session per sweep in `scheduled`, not one shared by all six.**
  `ctx.waitUntil` only registers a promise; what decides which session a query
  rides is where the promise was created. Sharing one session would couple six
  unrelated, delete-heavy sweeps to a single bookmark each of them keeps
  advancing, so every read would be forwarded to the primary anyway.

### Why concurrent requests cannot cross sessions

The whole design rests on two in-flight requests never sharing a store, and that
property is not owned by this code — it comes from how Effect schedules fibers.
Every service read runs through `dbQuery` → `Effect.promise`, so all queries
execute inside fiber drains. Since Effect 3.20, `SchedulerRunner.cached` keeps
runners in a `WeakMap` keyed by fiber, so each fiber's drain microtask is created
inside that fiber's own context. There are tests for exactly this, and they fail
under the older shared-runner behaviour — they are load-bearing, not
belt-and-braces.

The consequence to keep in mind when writing new code: a module-scope Effect
primitive that suspends one request's fiber and resumes it from another's (a
shared `Deferred`, `Semaphore`, `Queue`, `Effect.cached` value, or a forked
daemon fiber) is incompatible with this design. `@cire/api` uses none today, and
the wiki page says so.

## Workspaces affected

- `@cire/api` — new `src/db/d1-session.ts`; `createD1Db` widened to accept the
  narrow `D1QueryClient`; `src/index.ts` wraps `fetch` and each `scheduled` sweep.
- `wiki/` + `CLAUDE.md` — new `wiki/systems/d1-read-replication.md`, linked from
  the navigation table, `wiki/index.md`, `wiki/apps/cire-development.md`,
  `wiki/systems/d1-limits.md` and `wiki/systems/database-environments.md`.

## Issues

Two follow-ups filed from the review, both informational: `xchromo/osn-tracker`
#597 (P-I1) and #598 (P-I3).

## Decisions

- **`first-primary` Worker-wide**, rather than per-route constraints. Reasoning
  above; a route that ever genuinely wants `first-unconstrained` needs its own
  session and its own argument in writing, not a change to the shared constant.
- **A shim on the async context**, rather than threading a per-request Drizzle
  handle through the app. The app graph is isolate-cached by design (`aot: false`
  plus ~50 route factories closing over the handle); rebuilding it per request
  would cost far more than the round trips this saves.
- **`runInD1Session` calls `withSession` unguarded on purpose.** A D1 binding is
  required to have it, and a binding that does not — a stub, a wrong binding type
  — is a misconfiguration worth a loud failure rather than a silent lifetime of
  primary-only reads. That is deliberately different from the *missing session*
  fallback, which is silent because it is safe.
- **Ship before enabling replication**, so the mechanical change and the
  consistency-affecting change are separately revertible.

## Test plan

| Gate | Command | Result |
|---|---|---|
| Type check | `bunx --bun tsc --noEmit -p cire/api` | clean |
| Unit + route tests | `bun run --cwd cire/api test` | 1740 pass, 0 fail, 5413 expect() |
| Lint | `bunx --bun oxlint cire/api/src` | clean (pre-existing unused-var warnings in unrelated test files) |
| Format | `bunx --bun oxfmt --check cire/api/src` | clean |
| Changesets | `./scripts/validate-changesets.sh` | pass |
| Worker build | `bunx --bun wrangler deploy --dry-run` | builds |

New coverage, in three places:

- `src/db/d1-session.test.ts` — routing, fallback when no session is in scope,
  restoration after the scope ends, concurrent sessions kept apart, the same
  through the Effect fiber scheduler, and the constraint constant actually being
  what `withSession` receives. Then, against a real workerd-backed D1 through
  Miniflare: a Drizzle handle over the shim reads back its own writes, `.batch()`
  routes through the session too, the bookmark advances from null to a real
  position as the request queries, and — the premise of the whole substitution —
  Drizzle's D1 driver touches nothing but `prepare` and `batch`. `drizzle-orm` is
  on a caret range, so that last one fails loudly if a minor bump adds a call
  path a `D1DatabaseSession` cannot answer.
- `src/index.test.ts` — the entry points, through a recording D1 binding: one
  `first-primary` session per request; a distinct session per request through the
  shared isolate-cached app; two concurrent requests never seeing each other's
  session (read off the bound parameters, since Drizzle parameterises every value
  and the SQL text alone cannot say which request a query belonged to); six
  `waitUntil`ed sweeps to six sessions; and, on every one of those, that no query
  ever reaches the raw binding. That last assertion is the silent failure this
  change is most exposed to: a handle built over the raw binding would work
  perfectly and simply never use replication.
- `src/db/d1-integration.test.ts` — the suite's Drizzle handle now runs over the
  shim, so the existing integration tests exercise the production shape, plus one
  test that `claim.lookup` gives the same answer inside a session as outside.

Not covered by tests, and verified by reading: the behaviour once replicas exist.
There is no local way to run one.

## After merge

Enable read replication on `cire-db` (D1 → database → Settings → Read
replication, or the REST `PUT` documented on the wiki page) and re-measure
against the 35.7 ms/query baseline above. Billing is unchanged either way —
Cloudflare bills `rows_read`/`rows_written` with or without replicas.

### DIFF
--- a/.changeset/cire-d1-sessions-api.md
+++ b/.changeset/cire-d1-sessions-api.md
@@ -1,0 +1,36 @@
+---
+"@cire/api": patch
+---
+
+Route every request's D1 queries through one D1 session, so read replicas can serve them.
+
+Without a session, each `prepare()` is an independent query and D1 sends all of
+them to the primary database. cire's primary is in Oceania, so a request served
+from a European or North American colo pays the full round trip to Sydney *per
+query* — measured at ~36 ms each from a Sydney client against the OC primary,
+and far worse from further away. With replication on, a session pins the first
+query to the primary and lets every later query be served by any replica caught
+up to the bookmark the first one returned: N × distance collapses to one round
+trip plus N-1 local reads.
+
+- New `db/d1-session.ts`. The Elysia app graph is built once per isolate with
+  the Drizzle handle baked in and ~50 route factories closing over it, so a
+  per-request handle cannot be threaded explicitly. Instead the handle is built
+  over a stable client shim whose `prepare`/`batch` delegate to whichever
+  session is current on an `AsyncLocalStorage`, established per request in
+  `fetch` (and once around the six `scheduled` sweeps). Outside a session the
+  shim falls through to the raw binding — exactly today's behaviour — so losing
+  the context costs latency, never correctness.
+- `createD1Db` now takes the two-method `D1QueryClient` rather than a full
+  `D1Database`: those two methods (`prepare`, `batch`) are all Drizzle's D1
+  driver ever calls, and are exactly what a `D1DatabaseSession` exposes.
+- The constraint is `first-primary`, never `first-unconstrained`, so a request
+  always observes every write committed before it started. `routes/invite.ts`
+  serves a `no-store`, edit-sensitive payload precisely so organiser edits show
+  up on the guest invite's revalidation; one constraint Worker-wide keeps that
+  from depending on which route was reached.
+
+Safe to ship before replication is enabled: with no replicas a session behaves
+exactly as today. Tests cover the routing (including two interleaved requests
+kept apart across Effect's fiber scheduler, which is what every service query
+runs on) and a real workerd-backed D1 via Miniflare.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -70,6 +70,7 @@ One label is orthogonal to all of that: **`needs:decision`**, on both repos. It 
 |---|---|
 | Understand monorepo layout | `[[wiki/architecture/monorepo-structure]]` |
 | Understand DB environments (local bun:sqlite vs dev/staging/prod D1) | `[[wiki/systems/database-environments]]` |
+| Cut D1 latency with read replicas (the Sessions API, `first-primary`, one session per invocation, turning replication on) | `[[wiki/systems/d1-read-replication]]` |
 | Write new Effect service or Elysia route | `[[wiki/architecture/backend-patterns]]`, `[[wiki/architecture/schema-layers]]` |
 | Understand accounts, profiles, orgs | `[[wiki/systems/identity-model]]` |
 | Add or verify ARC S2S tokens | `[[wiki/systems/arc-tokens]]` |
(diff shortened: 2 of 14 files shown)


chav-aniket:
@claude how secure is having the database connection in the localstorage?

Issue #619, opened 2026-08-15
Verified identity V-M3: re-issue mDL claims as OSN SD-JWT verifiable credentials


chav-aniket:
Part of the V-M3 epic, #546 (mobile driver licence acceptance).

## What and why

OSN accepts state-issued mobile driver licences (mDL, CBOR/COSE format, #618) but will not issue them. Everything OSN issues is an SD-JWT VC — a selectively disclosed JSON Web Token verifiable credential (RFC 9901 plus the SD-JWT VC profile). So once an mDL presentation verifies, OSN stores the released fields as `verified_attributes` rows and presents them as SD-JWT VCs like any other attribute. Relying parties then see one credential format, whatever the source.

Nothing of verified identity exists in the code yet: no `verified_attributes`, `verification_runs` or `presentations` table in `osn/db/src/schema/index.ts`, and no `/identity/*` route under `osn/api/src/routes/`. It depends on the V-M0 SD-JWT VC issuance helper, which signs with its own ES256 key — same JSON Web Key Set (JWKS), different `kid` from the ARC service-to-service key.

## Proposal

- Map each released mDL field to an attribute kind (`given_name`, `family_name`, `dob`, `document_expires_at`, …) and work out `age_over_16` / `age_over_18` from `dob`, as #615 does for documents.
- Use the mDL's own age-over flags where the wallet releases them without `dob`.
- Record provenance: provider `mdl`, the issuing state, and the verification run, so Settings can show "Verified … via NSW digital driver licence".

## Done when

- A verified NSW or Queensland mDL presentation writes `verified_attributes` rows tagged with the mDL provider and issuing state.
- A presentation of those attributes is an SD-JWT VC signed with the VC key and verifiable against `/.well-known/jwks.json`.
- No CBOR or COSE value reaches a relying party.

## Files

- `osn/api/src/` — the mDL-to-attribute mapping (new)
- `osn/db/src/schema/index.ts` — `verified_attributes` (planned in V-M0)
- Design: `wiki/osn/verified-identity.md` §Cryptography choice ("Not chosen for v1: mdoc/COSE issuance"); the ARC service-to-service key the VC key must stay apart from is described in `wiki/osn/arc-tokens.md`

Issue #612, opened 2026-08-15
Verified Identity V-M2: register for the Australian DVS and show its approved consent statement


chav-aniket:
## What and why

The Document Verification Service (DVS) is the Australian Government service that answers yes or no to "does this licence or passport match government records". Before OSN, through a gateway provider, may call it, OSN has to register with the Department of Home Affairs and show users a DVS-approved consent statement. The Australian Privacy Act 1988 and Australian Privacy Principle 11 also apply once OSN handles this data.

Part of the Verified Identity V-M2 milestone (Australian document verification: DVS + selfie + face-match), epic #545. The design is in `wiki/osn/verified-identity.md`. Nothing from that design has shipped yet; the schema and credential issuer come from the V-M0 foundations epic, #543.

## Proposal

- **Paperwork (owner task):** complete the Department of Home Affairs DVS registration.
- **Consent screen:** show the DVS-approved consent statement, word for word, before **every** DVS call. The user must accept it for that call.
- **Consent record:** store a record of each acceptance (account, time, statement version) and keep it.

## Done when

- The DVS registration is approved.
- The V-M2 flow cannot reach the gateway provider (#613) without a stored consent record for that run.
- The consent-record store appears in `wiki/compliance/data-map.md` with its retention rule in `wiki/compliance/retention.md`.

## Files

- `wiki/osn/verified-identity.md` — §Compliance posture
- `wiki/compliance/data-map.md`, `wiki/compliance/retention.md`

PR #904, opened 2026-09-06
chore: give all six Astro apps a bundle-size guard

chav-aniket:
## Summary

`cire/invites` was the only Astro app in the repo with a bundle-size guard. This
branch gives all six one, and adds the second half of tracker #287 that a file
move never fixed: a check that no test file sits routable under any app's
`src/pages`.

The tracker issue's premise needed correcting first. It assumed the other five
apps "build the same way" — they do not. `cire/invites` is the only
`output: "server"` app; `cire/host`, `cire/vendor`, `cire/landing`,
`osn/landing` and `pulse/landing` are `output: "static"` with no `dist/server`
and no Worker at all. So the guard takes a mode: `worker` measures `dist/server`
as before, `static` measures `dist/_astro/*.js` and `*.css` only.

- Every threshold lives in `scripts/bundle-size-budgets.txt` and nowhere else.
  An app with no row is an error, not a skip — a guard that silently covers
  nothing is the failure this family exists to prevent.
- Headroom is sized *smaller* than the mistake being caught (a motion-class
  dependency, 21261 bytes gzip minified), so one such library trips every app's
  guard rather than fitting inside its budget.
- The guard runs from each app's `build` script **and** as its own CI/deploy
  step. Those are not redundant: a Turborepo cache replay of `build` reproduces
  the logs without executing the chained script, so the standalone step is the
  only thing that checks a cache-hit run.

## Workspaces affected

`@cire/invites`, `@cire/host`, `@cire/vendor`, `@cire/landing`, `@osn/landing`,
`@pulse/landing`, plus `scripts/`, `.github/workflows/` and `wiki/`.

Two changesets, split as the repo requires: `@cire/*` packages are version-less
and must never share one with a versioned package, and `@osn/landing` /
`@pulse/landing` both carry a `version`. `scripts/changeset-required.sh` says
`required` for this diff — the deleted `cire/invites/scripts/guard-ssr-size.sh`
path is on no allowlist — and `scripts/validate-changesets.sh` passes.

## Issues

**Closes**

| Issue | What |
|---|---|
| xchromo/osn-tracker#619 | `P-I8` |
| xchromo/osn-tracker#635 | `T-M1` — the moved guard now has a CLI test |

Closes xchromo/osn-tracker#619
Closes xchromo/osn-tracker#635

**Opened**

| Issue | What |
|---|---|
| xchromo/osn-tracker#636 | `P-I1` |

### Correct the issue's premise: two modes, not one measurement — `P-I8`

- **Issue** — tracker #619 assumed six apps that build alike. Five of them are
  `output: "static"`: no Worker, no `dist/server`, deployed to Pages. "SSR Worker
  bundle size" is not a number that exists for them.
- **Why** — generalising the existing script unchanged would have produced five
  guards measuring a directory that never exists, which fails or, worse, gets
  quietly softened into a skip.
- **Solution** — a `worker` mode (unchanged behaviour, `cire/invites` only) and a
  `static` mode measuring `dist/_astro/*.js` + `*.css`. Not all of `dist`:
  `astro build` downloads Google font binaries into `dist/_astro/` with hashed
  names, and those bytes move when Google's metadata moves with no repo change;
  `dist` also holds one HTML file per page, so it grows with ordinary content.
- **Rationale** — `*.js` + `*.css` is the part that regresses when a library
  wanders in, which is the only thing this guard is for. The worker mode is an
  exclude-set and the static mode an allowlist, deliberately and for different
  reasons — both are argued in the script's own comment.

### One committed table, not a threshold per call site

- **Issue** — the first draft took the threshold as a command-line argument. Each
  number then appeared three to eight times: the package's `build` script, the
  `ci.yml` step, two `deploy.yml` jobs, and the wiki.
- **Why** — re-baselining after an intentional bundle change means finding every
  copy. Miss one and CI and the build script disagree silently, with the lower
  number winning and nothing saying why: the guard then either cries wolf or
  stops guarding.
- **Solution** — `scripts/bundle-size-budgets.txt` holds one row per app
  (`<package-dir> <mode> <threshold>`, with the measured reading in a trailing
  comment). `guard-bundle-size.sh <pkg>` looks the row up; `--all` walks the
  table. There is deliberately no way to pass a threshold on the command line.
- **Rationale** — every threshold now appears exactly once outside the wiki,
  verified by grep. A package with no row is a hard error rather than a skip.

### Headroom smaller than the mistake, and flat rather than proportional

- **Issue** — how much room each app gets above its measured baseline.
- **Why** — the parent branch had to fix a threshold whose headroom was *larger*
  than the library it was meant to catch, which made it structurally unable to
  fire on a single-library mistake.
- **Solution** — a flat 11683 bytes for every app, against measured baselines of
  15182 (`osn/landing`) to 212832 (`cire/host`).
- **Rationale** — a dependency's byte cost does not scale with the size of the
  app it lands in, so headroom sized to "less than one bad dependency" is
  correctly a constant, not a percentage. 11683 is well under the 21261-byte
  motion-class figure on every row, so that class of mistake trips the guard on
  the smallest app and the largest alike.

### The `src/pages` check fails closed — `T-U3`

- **Issue** — the route checker resolved each app through a relative path and
  treated an unreadable directory as "nothing to check". The review reproduced
  the consequence: move the script so the join breaks, and it prints its success
  line and exits 0 while checking zero directories.
- **Why** — a guard that passes vacuously is indistinguishable in CI output from
  a genuinely clean repo, and is functionally identical to no guard.
- **Solution** — a configured app whose `src/pages` is missing is now a hard
  error naming the path it tried. All six apps have one today, so a missing one
  is always a bug. A `src/pages` that exists and is empty still passes.
- **Rationale** — verified both ways after the fix: the real run passes, and the
  broken-resolution scenario now exits 1 with a message naming the path.

### The budgets parser must not glob — `S-M1`

- **Issue** — rows were split with an unquoted array assignment. Bash performs
  pathname expansion there, against the caller's working directory.
- **Why** — a row containing a `*` would expand to whatever files sit beside the
  caller, and the guard would then measure an unrelated directory and pass. No
  live row contains a glob character, so nothing was affected, but the failure
  mode is the vacuous pass again.
- **Solution** — `read -r`, which word-splits without globbing, at both parse
  sites; a fourth field is now caught rather than folded into the third.
- **Rationale** — confirmed by hand that a `* worker 999999999` row now fails
  closed instead of measuring a neighbouring directory.

### `osn/landing` and `pulse/landing` get no extra deploy step

- **Issue** — whether every app needs a standalone guard step in its deploy job.
- **Why** — the standalone step exists because a Turborepo cache replay skips the
  chained `build` invocation.
- **Solution** — the four cire deploy jobs build with `bun run build --filter=…`
  (turbo, cacheable) and get one. `deploy-osn-pulse-landing.yml` runs
  `bun run --cwd <dir> build`, which calls the package script directly and
  bypasses turbo entirely, so the chained invocation always executes and a second
  step would be noise.
- **Rationale** — the rule is "guard every path that can ship an artefact", not
  "add a step everywhere".

**Out of scope** — xchromo/osn-tracker#636 — `P-I1`. Also unfixed and unclaimed:
tracker #612 stays open — moving this script out of `cire/invites/scripts/` puts
it under CODEOWNERS, but `pulse/api/scripts/`, `osn/api/scripts/` and
`cire/api/scripts/` are still outside every pattern, which is what that issue is
actually about. Separately, `osn/landing` and `pulse/landing` use
`fontProviders.google()` but are not covered by `ci.yml`'s Astro font-metadata
cache step, so they fetch from Google uncached on every CI run — pre-existing,
noticed while measuring, untouched here.

## Test plan

| Gate | Command | Result |
|---|---|---|
| Lint | `bun run lint` | pass |
| Format | `bun run fmt:check` | pass (plus `oxfmt scripts --check` by hand — the root glob does not scan `scripts/`) |
| Type check | `bun run check` | pass — 37/37 tasks |
| Build | `bun run build` | pass — 13/13 |
| Script tests | `bun run test:scripts` | pass — 143 tests, 278 assertions, 13 files |
| Changesets | `./scripts/validate-changesets.sh` | pass |
| Changeset needed? | `git diff --name-only perf/cire-invites-ssr-bundle...HEAD \| ./scripts/changeset-required.sh` | `required` — two are present |
| All guards | `./scripts/guard-bundle-size.sh --all` | pass, all six |
| Route check | `bun run scripts/check-astro-test-routes.ts` | pass |

Measured readings behind the thresholds, from one clean build with no `PUBLIC_*`
set and nothing else building:

| App | Mode | Measured | Files | Threshold |
|---|---|---:|---:|---:|
| `cire/invites` | worker | 163334 | 33 | 175000 |
| `cire/host` | static | 212832 | 33 | 224515 |
| `cire/vendor` | static | 69961 | 10 | 81644 |
| `cire/landing` | static | 177641 | 9 | 189324 |
| `osn/landing` | static | 15182 | 5 | 26865 |
| `pulse/landing` | static | 16683 | 5 | 28366 |

Exercised by hand, because a guard that only ever passes proves nothing:

| Check | Result |
|---|---|
| a planted `cire/vendor/src/pages/zz-drift.test.ts` | exit 1, names the file |
| the same file `_`-prefixed | exit 0 |
| route checker with its relative join broken | exit 1, names the path it tried (was: exit 0, "no violations") |
| a `* worker 999999999` budgets row | fails closed (was: would measure a neighbouring directory and pass) |
| `guard-bundle-size.sh` from inside a package (`.`) | resolves the right row |
| `guard-bundle-size.sh osn/api` (no row) | exit 1, "no budget recorded" |

Not verified here: that a real CI run hits the cache-replay path the standalone
deploy steps defend against. It was reproduced locally — a second `bun run build`
came back FULL TURBO with none of the guards' output — but the first CI run on
this PR is the real check. `cire/invites`'s reading drifts by 1-2 bytes between
builds (Astro regenerates a props-encryption key each time); the headroom absorbs
it.


chav-aniket:
@claude double check the guard script for conciseness and accuracy

chav-aniket:
> **Claude finished @chav-aniket's task in 3m 30s** —— [View job](https://github.com/xchromo/osn/actions/runs/34039878209)
> 
> ---
> ### Reviewing `scripts/guard-bundle-size.sh` for conciseness and accuracy
> 
> - [x] Gather context (read PR diff, guard script, budgets table, tests)
> - [x] Review script for correctness (worker vs static modes, parsing, error handling)
> - [x] Review script for conciseness (redundancy, dead code, simplification opportunities)
> - [x] Post findings
> 
> Read `scripts/guard-bundle-size.sh` in full, cross-checked it against `scripts/bundle-size-budgets.txt`, both call sites (`cire/host/package.json` etc., `ci.yml`, `deploy.yml`), and `scripts/tests/guard-bundle-size.cli.test.ts`'s own "one case per decision point" list. Also looked at the sibling `check-astro-test-routes.ts` since it's part of the same PR and same "fail closed, never vacuously pass" theme.
> 
> **Overall: the script is accurate.** The worker/static split, the `dist/client/*.map` tripwire, the `read -r` (not an unquoted array) glob-safety fix, and the fail-closed checks (`any_record`, per-row `count -eq 0`, `validate_budgets_file`) all do what their comments say, and I confirmed the failure-message text against what the tests assert. The heavy commenting is proportionate for a script whose whole point is "the reasoning must survive the next re-baseline" — I wouldn't cut it for concision.
> 
> Two small gaps, both edge cases rather than bugs in the guard's normal operation:
> 
> 1. **`--all` + a row whose package directory doesn't exist skips the fail-closed message.** `run_guard` (`scripts/guard-bundle-size.sh:129-148`) does `cd "$pkg_dir"` with no existence check first when called from `run_all` (`scripts/guard-bundle-size.sh:295-307` builds `pkg_dir` as `"$BUDGETS_ROOT/$REC_PKG"` and passes it straight through). `lookup_and_run`'s single-package path checks existence first via `resolve_label` and emits `::error::guard-bundle-size.sh: package directory '...' does not exist.` (tested — `scripts/tests/guard-bundle-size.cli.test.ts:325-341`, the `T-U1` case) — but that check never runs for `--all`. A bad/stale row in `bundle-size-budgets.txt` under `--all` would instead surface as a raw `cd: no such file or directory` from bash, without the `::error::` prefix or the polished explanation every other failure path gets. It still exits non-zero (`any_fail=1` still gets set), so this isn't a silent pass — just an inconsistent, untested message path. No test in the suite's own enumerated list (`scripts/tests/guard-bundle-size.cli.test.ts:18-33`) covers "`--all` with a missing directory," only the single-package case does.
> 
>    [Fix this →](https://claude.ai/code?q=In%20scripts%2Fguard-bundle-size.sh%2C%20the%20run_all()%20function%20(around%20line%20295-307)%20calls%20run_guard%20with%20a%20pkg_dir%20built%20from%20BUDGETS_ROOT%2FREC_PKG%20without%20first%20checking%20the%20directory%20exists%2C%20unlike%20lookup_and_run()%20which%20uses%20resolve_label()%20to%20check%20existence%20and%20emit%20a%20polished%20%3A%3Aerror%3A%3A%20message.%20Add%20the%20same%20existence%20check%20in%20run_all()%20before%20calling%20run_guard%2C%20emitting%20a%20consistent%20%3A%3Aerror%3A%3A%20message%20naming%20the%20missing%20directory%2C%20and%20add%20a%20CLI%20test%20in%20scripts%2Ftests%2Fguard-bundle-size.cli.test.ts%20covering%20--all%20with%20a%20budgets%20row%20whose%20package%20directory%20does%20not%20exist.%20Branch%3A%20chore%2Fastro-bundle-guards&repo=xchromo/osn)
> 
> 2. **`check-astro-test-routes.ts`'s test-only override can silently check zero apps.** `ASTRO_APPS = Bun.env.ASTRO_TEST_ROUTE_APPS?.split(",").filter(Boolean) ?? DEFAULT_ASTRO_APPS` (`scripts/check-astro-test-routes.ts:75`) — `??` only falls back on `undefined`/`null`. If that env var were ever set to `""` (or something that `.filter(Boolean)` reduces to nothing), `ASTRO_APPS` becomes `[]`, the `for` loop body never runs, `failed` stays `false`, and the script prints the success banner (`scripts/check-astro-test-routes.ts:128-130`) having checked nothing — the exact "vacuous pass" class of bug this PR's own `T-U3` fix and `guard-bundle-size.sh`'s `any_record` check (`scripts/guard-bundle-size.sh:309-312`) exist to close, just not applied here. Low real-world severity: in production `ci.yml` calls the script with no env var set, so `DEFAULT_ASTRO_APPS` (6 apps) is always used today, and only the CLI test harness ever sets `ASTRO_TEST_ROUTE_APPS`. Worth a one-line `if (ASTRO_APPS.length === 0)` guard for consistency with the sibling script's philosophy, but not urgent.
> 
> Minor conciseness notes, neither worth a fix on their own: `lookup_and_run`/`run_all` compute `stripped` for the blank-line check and then call `parse_record("$line")`, which re-derives the same `${1%%#*}` strip internally — passing `$stripped` instead would drop the redundant re-strip. And the `case "$mode" in worker|static|*) ...)` default arm in `run_guard` (`scripts/guard-bundle-size.sh:141-148`) is unreachable today since `validate_budgets_file` already rejects any other mode before either call site — the script's own comment says as much and frames it as deliberate defense-in-depth, which is a reasonable call to leave as-is.
> 
> No changes made — this was requested as a review, not an implementation.
> 
> [Branch: chore/astro-bundle-guards](https://github.com/xchromo/osn/tree/chore/astro-bundle-guards)

@claude apply fixes

Issue #895, opened 2026-09-06
Epic: migrate the monorepo from Effect 3.22 to Effect 4.0


chav-aniket:
Effect (the `effect` npm package, the functional-effect-system library this repo's backends are built on) is pinned at `^3.22.1` in 13 workspace packages. Effect 4.0 is a major release with a large breaking surface. This epic tracks the bump.

361 TypeScript files import `effect` or an `@effect/*` package — `cire/api` (144), `osn/api` (96), `pulse/api` (65), `zap/api` (16), `osn/client`, the four `*/db` packages, and `shared/crypto`, `shared/email`, `shared/redis`, `shared/observability`, `shared/db-utils`. 793 tests run through `@effect/vitest`'s `it.effect`.

Full plan: **`wiki/runbooks/effect-v4-migration.md`**.

## Merge strategy: the `effect-v4` integration branch (decided 2026-09-06)

**The branch exists**, cut from `main` at `851df71`.

```
main ──────────────────────────────────────────────► (one green PR at the end)
  └── effect-v4 ◄── phase 1 ◄── phase 2 ◄── … ◄── phase 6
```

This is settled because phases 1–4 **cannot pass CI** (see the spike findings below) and `main` requires it. An integration branch is the only option that keeps both a reviewable, phase-sized diff and an always-green `main`; reviewing a ~1300-site diff commit-by-commit in one PR gives up the first, and waiving the gate on a stack merging to `main` gives up the second.

- **Every phase PR bases on `effect-v4`**, never `main`: `gh pr create --base effect-v4`. Follow `wiki/conventions/stacked-prs.md` for the rest.
- **Phases 1–4 will show red CI, by design.** `ci.yml` runs on every `pull_request` regardless of base. Merge them into `effect-v4` with the red understood, and **never via auto-merge** — it would wait on a check that cannot go green. Phase 5 is the first PR whose CI can pass.
- **Rebase `effect-v4` onto `main` regularly.** Every Effect-typed package is in scope, so conflicts are the norm.
- **The final merge to `main` is one green PR** with the full suite, both D1 tiers and the dev-tier smoke behind it (#903).

Verified against the workflows on 2026-09-06:

| Workflow | Trigger | Effect on `effect-v4` |
| --- | --- | --- |
| `deploy.yml` | `push` to `main` only | **No deploy.** Nothing reaches the dev tier until the final merge |
[...]

PR #907, opened 2026-09-06
docs(effect): install the official Effect skills and plan the v4 migration

chav-aniket:
## Summary

Everything the Effect v4 migration needs in place before its code changes start: the official skills, the guards around them, and the costed plan. **No application code changes** — this is docs and tooling, green on its own, and it merges to `main` ahead of the migration stack.

It exists as a separate PR because `effect-v4` was cut from `main` and carries none of it — not the skills, not the `.gitignore` entry for `.repos/`, not the CODEOWNERS coverage, not the changeset allowlist, not the runbook. Phase 1 (#906) hit that immediately: it staged two embedded git repositories by accident and had to duplicate the `.gitignore` entry to get past it. Once this lands, `effect-v4` rebases onto `main` and picks all of it up, and the duplicated hunk resolves itself.

**The official skills.** `npx skills add Effect-TS/skills` installs `effect-ts` (which points agents at `node_modules/effect/AGENTS.md`, so guidance tracks the installed version) and `effect-v3-to-v4` (the migration procedure). The migration one refuses to guess: every rename resolves from upstream's generated `migration/v3-to-v4.md`, searched a symbol at a time, and it forbids a v3-compat shim, an `any`/`as` to silence a post-migration error, and any invented replacement API.

**The plan.** `wiki/runbooks/effect-v4-migration.md` costs the bump against this repo — 361 files, 793 `it.effect` tests, ~1300 edit sites — and carries what the phase-0 spike and phases 1–2 have already corrected in it.

## Workspaces affected

None. No workspace package changes, so no changeset is required — `scripts/changeset-required.sh` returns `skip`, via the `.claude/*`, `.agents/*`, `scripts/*` and `wiki/*` allowlist entries.

Files touched: `.agents/skills/`, `.claude/skills/` (symlinks), `skills-lock.json`, `.github/CODEOWNERS`, `.gitignore`, `CLAUDE.md`, `scripts/changeset-required.sh` (+ its test), `wiki/index.md`, `wiki/runbooks/effect-v4-migration.md`.

## Issues

Sets up #895 and its phases #896–#903. Closes none of them.

No findings filed — nothing here touches application code.

### Third-party skills live in `.agents/`, not `.claude/skills/` directly

`CLAUDE.md` said a procedure lives in `.claude/skills/<name>/SKILL.md` "and nowhere else". The official installer writes `.agents/skills/<name>/` and symlinks `.claude/skills/<name>` at it, with `skills-lock.json` naming the source repo and content-hashing each file.

Kept the installer's layout rather than copying the text in, so `npx skills update` can move the pin, and amended the `CLAUDE.md` rule to name this as the one exception. Vendoring by hand would have matched the existing `tools/oxlint/anti-slop` precedent but given up the update path.

### CODEOWNERS covers the skills and the lock

A skill runs with full agent permissions, so third-party skill text gets the same human review as the vendored oxlint plugin. `/.agents/**`, `/.claude/**` and `/skills-lock.json` are now owned.

### `.agents/*` and `skills-lock.json` added to the changeset allowlist

Same reasoning the file already applies to `.claude/*`: agent instructions ship in no package. Two fixtures added to `scripts/tests/changeset-required.test.sh`; the suite passes 21/21.

### `.repos/` is gitignored here, but the oxlint exclusion is **not** in this PR

The migration skill clones the Effect repo into `.repos/` as its reference, and `.gitignore` keeps it out of git.

I originally also excluded it in `oxlintrc.json` and had to take that back out (`72ad216`) — it made this PR fail `Require changeset`, and that was my error: I ran `scripts/changeset-required.sh` on the diff, then added `oxlintrc.json` in a later commit and never re-ran it. Root files require a changeset; the allowlist covers `.gitignore` and root `*.md`, not `oxlintrc.json`.

Adding a changeset would have been the wrong fix — it means naming a package and writing a changelog entry for something that ships in no package, which is exactly what the allowlist exists to prevent. Adding `oxlintrc.json` *to* the allowlist may well be right on the script's own stated logic (it versions nothing, and does not change what deployed packages build from), but that is a change to a guard and belongs in its own PR with an owner's review.

**Consequence, worth knowing:** until the migration stack lands, running the `effect-v3-to-v4` skill from a branch off `main` produces oxlint warnings from the ~16k files under `.repos/`. Noise in a local lint run, not a correctness problem. The identical exclusion already sits on the phase-1 branch (#906), which carries a changeset, so it reaches `main` with the migration.

### The plan diverges from the skill on what counts as done

The skill's done condition is a clean type-check, and it says explicitly that tests are *not* a gate. The runbook overrides that: **tests and the dev tier are the gate here.** The phase-0 spike produced the worked example — `Effect.either` → `Effect.result` also changes the result's `_tag` from `"Right"`/`"Left"` to `"Success"`/`"Failure"`, and because `toBe` takes `any`, six assertions type-checked clean while comparing a v4 tag against a v3 string. The skill's own done condition would have called that migration finished.

### Tessl Skill Review suggestions are deliberately not applied

Both SKILL.md files passed (81% and 92%, 16/16 checks each). Its suggestions are description improvements, and these files are **vendored upstream and content-hashed in `skills-lock.json`**. Editing them would break the hash so `npx skills update` cannot move the pin cleanly, fork us from upstream so every future Effect-TS improvement arrives as a conflict, and mean the text reviewed at install time is not the text in the tree. Upstream is where those suggestions belong.

## Test plan

- **`bun run lint`** — exit 0.
- **`bun run fmt:check`** — passes, 1444 files.
- **`bun run check`** — 37/37 workspaces successful.
- **`scripts/tests/changeset-required.test.sh`** — 21/21 pass, including the two new fixtures for `.agents/*` + `skills-lock.json`.
- **`scripts/changeset-required.sh`** on this branch's own diff — returns `skip`.
- **Skill install verified** — both skills resolve through the `.claude/skills/` symlinks and appear in the session's skill list; `git ls-files -s` shows mode `120000` for the two symlinks and `100644` for the real files.
- **CI green on `3c0d4d9`** — all 11 checks pass or skip (`swift test` skipped, no Swift changes), `mergeable_state: clean`.

One caveat worth naming: switching between this branch and a migration phase branch requires `bun install`, since their lockfiles pin different Effect majors. Linting this branch against a phase branch's `node_modules` reports false failures — that happened once while preparing this PR.


chav-aniket:
## `Require changeset` — fixed, and it was my mistake

`oxlintrc.json` was the sole cause. I ran `scripts/changeset-required.sh` on the diff, then added the oxlint change in a *later* commit and never re-ran it — so this PR's body claim that the check returns `skip` was true when I checked and false by the time I opened it. Root files require a changeset; the allowlist covers `.gitignore` and root `*.md`, not `oxlintrc.json`.

Dropped the oxlint change in `72ad216`. The diff is docs and agent config again, and `changeset-required.sh` returns `skip`.

**Adding a changeset would have been the wrong fix** — it means naming a package and writing a changelog entry for something that ships in no package, which is precisely what the allowlist exists to prevent. Adding `oxlintrc.json` *to* the allowlist may well be correct on the script's own stated logic (it versions nothing, and it does not change what deployed packages build from), but that is a change to a guard and belongs in a PR about the guard with an owner's review — not smuggled into a docs branch.

Nothing is lost: the identical `.repos` exclusion is already on the `effect-v4` phase-1 branch (#906), which carries a changeset, so it reaches `main` when the migration lands. Until then, running the `effect-v3-to-v4` skill from a branch off `main` produces oxlint warnings from the cloned Effect repo under `.repos/`. That is noise in a local lint run, not a correctness problem.

## Tessl Skill Review — no action, and deliberately so

Both files passed (81% and 92%, 16/16 validation checks each). The suggestions are all about the *descriptions* — broader trigger terms, more concrete setup actions, a verification step.

**I am not applying them, because these files are vendored upstream and editing them is the one thing this PR is structured to avoid.** `.agents/skills/effect-ts/SKILL.md` and `.agents/skills/effect-v3-to-v4/SKILL.md` come from `Effect-TS/skills` via `npx skills add`, and `skills-lock.json` content-hashes each one. Editing them locally would:

- break the hash, so `npx skills update` can no longer move the pin cleanly;
- fork us from upstream, so every future improvement Effect-TS makes to the migration procedure arrives as a conflict;
- and mean the text a human reviewed at install time is no longer the text in the tree.

That trade-off is the reason `CLAUDE.md` now names third-party skills as the one exception to the "skills live in `.claude/skills/`" rule. If the suggestions are worth having, upstream is where they belong.


---

Issue #900, opened 2026-09-06
Effect v4 phase 3: rewrite Logger/LogLevel and the @effect/opentelemetry module renames


chav-aniket:
Part of the Effect 3.22 → 4.0 migration (#895). Full plan: `wiki/runbooks/effect-v4-migration.md`. Follows #899.

> [!important] Base this PR on `effect-v4`, not `main`
> `gh pr create --base effect-v4`. **CI will be red, by design** — per #896 the tree does not type-check until #902. Do not enable auto-merge.

**Invoke the `effect-v3-to-v4` skill and resolve every symbol from upstream's `migration/v3-to-v4.md`.** Tables below are sizing, checked against upstream on 2026-09-06; the reference wins where they differ.

Effect 4 rewrote `Logger`, and `LogLevel` is no longer a set of branded constructors. 28 sites across 12 files. This is the smallest phase by line count and the one with the highest chance of breaking something invisibly.

## Mapping

| v3 | v4 |
| --- | --- |
| `Logger.pretty` | `Logger.layer([Logger.consolePretty(), Logger.tracerLogger])` |
| `Logger.prettyLogger` | `Logger.consolePretty` |
| `Logger.prettyLoggerDefault` | `Logger.consolePretty()` — the prebuilt singleton became a constructor call |
| `Logger.replace` | `Logger.layer([…desiredLoggers])` |
| `Logger.jsonLogger` | `Logger.formatJson` |
| `Logger.withMinimumLogLevel` | `Effect.provideService(effect, References.MinimumLogLevel, level)` |
| `Logger.minimumLogLevel` | `Layer.succeed(References.MinimumLogLevel, level)` |
| `LogLevel.Debug` / `Info` / `Warning` / `Error` / `Fatal` / `Trace` / `All` | The string literals `"Debug"`, `"Info"`, … — v4 levels are literals used as both value and singleton type |

## Two regressions that type-check clean

**1. `Logger.layer` replaces the whole active logger set.** A migration that writes `Logger.layer([Logger.consolePretty()])` compiles, runs, and logs normally — while silently dropping log-to-span correlation, because `Logger.tracerLogger` is no longer in the set. Upstream is explicit: include it to preserve the v3 built-in layer behaviour. Every `Logger.layer` call in this repo must list every logger that has to stay active.

Verify by checking that a log line emitted inside a span still shows up attached to that span — not just that logs arrive.

[...]

PR #908, opened 2026-09-06
feat(effect): apply the v4 renames and rework Cause/Runtime error shaping

### LINKED ISSUE #899: Effect v4 phase 2: mechanical renames, removed APIs, and the ManagedRuntime/Runtime rework
Part of the Effect 3.22 → 4.0 migration (#895). Full plan: `wiki/runbooks/effect-v4-migration.md`. Follows #897 + #898.

> `gh pr create --base effect-v4`. The integration branch was cut from `main` at `851df71`. **CI on this PR will be red, by design** — per #896 the tree does not type-check until the `Schema` work (#902) lands. Do not enable auto-merge; it would wait on a check that cannot go green.

**Invoke the `effect-v3-to-v4` skill and resolve every symbol from upstream's `migration/v3-to-v4.md`.** The tables below are sizing, checked against upstream on 2026-09-06; where they disagree with the reference, the reference wins. No `v3-compat.ts`, no `any`/`as` to silence an error, no invented APIs.

This phase clears everything that is not `Schema` or `Logger` (#900–#902).

## Renames

| v3 | v4 | Sites | Files |
| --- | --- | ---: | ---: |
| `Effect.catchAll` | `Effect.catch` | 42 | 24 |
| `Effect.catchAllDefect` | `Effect.catchDefect` | 94 | 20 |
| `Effect.catchAllCause` | `Effect.catchCause` | 1 | 1 |
| `Effect.either` | `Effect.result` | 71 | 13 |
| `Effect.forkDaemon` | `Effect.forkDetach` | 7 | 5 |
| `Effect.zipRight` | `Effect.andThen` | 7 | 3 |
| `Effect.dieMessage` | `Effect.die(new Error(msg))` | 2 | 2 |
| `Layer.scoped` | `Layer.effect` | 6 | 4 |
| `Cause.failureOption` | `Cause.findErrorOption` | 4 | 4 |

`Effect.catchAll` → `Effect.catch` needs a word-boundary match, or it also rewrites `catchAllDefect` and `catchAllCause`. Do those two first. Read `migration/error-handling.md` for the `catch*` family and `migration/cause.md` for the flattened `Cause`.

## `Either` module → `Result`, and the trap the spike hit

Effect 4 dropped `Either`. `isLeft`→`Result.isFailure`, `isRight`→`Result.isSuccess`, `left`→`Result.fail`, `right`→`Result.succeed`. All four files are tests, and they pair with `Effect.either` → `Effect.result` above:

- `cire/api/tests/services/changes.test.ts`
- `shared/email/tests/cloudflare.test.ts`
- `shared/email/tests/resend.test.ts`
- `zap/api/tests/services/messages.test.ts`

> [!warning] `_tag` compared as a string is invisible to the compiler
> `Either`'s variants are tagged `"Right"` / `"Left"`; `Result`'s are `"Success"` / `"Failure"`. Any assertion of the form `expect(x._tag).toBe("Right")` still **type-checks clean** after the rename — `toBe` takes `any` — while being false at runtime. The phase-0 spike hit exactly this in `shared/crypto`: six assertions passed the compiler and failed the test run.
>
> **Grep for `_tag).toBe(` across the repo before finishing this phase** and convert every `"Right"`/`"Left"` literal. The type-check will not tell you.

## Removals — each needs a judgement call

`Layer.scoped` → `Layer.effect`: scoped acquisition merged in, and `Layer.effect` now supplies and excludes the layer `Scope`. Read `migration/scope.md`. Sites: `shared/redis/src/service.ts`, `shared/redis/src/ioredis.ts`, `cire/api/tests/services/import.test.ts`, `cire/api/tests/db/test-layer.ts`.

**The two `shared/redis` ones manage a real connection lifecycle, and a `Layer.effect` that drops the scope type-checks fine.** Leaked Redis connections do not crash anything obvious: the limiters fail closed (`wiki/systems/rate-limiting.md`), so the symptom is rejected requests. The tests have to catch this, not the compiler.

## `ManagedRuntime` and `Runtime` — read `migration/runtime.md`

**`ManagedRuntime` no longer extends `Effect`.** `runtimeEffect` / `runtime` became `contextEffect` / `context`; `ManagedRuntime.Context` became `ManagedRuntime.Services<T>`; `make` accepts `{ memoMap }`; `ManagedRuntimeUnify` and `TypeId` are gone (use `isManagedRuntime`).

**`Runtime<R>` is removed entirely.** `Context<R>` replaces it; run functions live directly on `Effect`, and the `Runtime` module is reduced to process-lifecycle utilities. This is also where `Runtime.isFiberFailure` and `Runtime.FiberFailureCauseId` went — the reference maps both to `none`, with the guidance to use an `Exit`-returning runner and inspect the `Exit` or `Cause`.

| File | What |
| --- | --- |
| `osn/api/src/lib/route-runtime.ts` | `AppRuntime` type, `makeAppRunner`, the injected-vs-fallback runtime logic |
| `osn/api/src/build-deps.ts:404` | `ManagedRuntime.make(Layer.merge(…))` at boot |
| `osn/api/src/app.ts:84` | the shared runtime on `AppDeps` |
| `osn/api/src/lib/safe-error.ts` | `Runtime.isFiberFailure` — the error path every route returns through |
| `osn/api/src/lib/grant-failure.ts` | same |
| `osn/api/tests/lib/route-runtime.test.ts`, `tests/lib/grant-failure.test.ts` | the tests for both |

`safe-error.ts` and `grant-failure.ts` decide what a client sees when something fails. Rewrite them by hand, not by script, and **confirm no internal error detail starts leaking into a response that previously returned a generic failure.**

## Not in scope

`Data.TaggedError` (165 sites) has no entry in the migration reference — unchanged. Leave it alone, and do not opportunistically convert it to `Schema.TaggedError`; that is a separate decision.

Layer memoization also changed (now shared across `Effect.provide` calls), which makes `CLAUDE.md`'s stated rationale for the shared-`ManagedRuntime` pattern inaccurate under v4. **The pattern stands — do not start using per-request `Effect.provide` on the strength of it.** Only the explanation changes, and #903 rewrites that row.

## Done when

The error inventory has shrunk to only the `Schema` and `Logger` errors that #900–#902 fix, and no new class of error has appeared. **A green CI is not available at this phase and is not the gate** — verify against the expected remaining set, and run the tests of any package that is fully migrated by this point.

chav-aniket:
## Summary

Phase 2 of the Effect v4 migration (#895), closing #899. Stacked on #906 (phase 1) — base is `claude/effect-v4-phase-1`, so this PR's diff shows only phase 2's work. It follows #906 into `effect-v4`.

Every replacement traced to upstream's generated reference or to v4 source. Nothing guessed, no compat shim, no `any`/`as` added — one pre-existing cast was *removed*.

**Straight renames:** `catchAllDefect` → `catchDefect` (94), `catchAll` → `catch` (42), `either` → `result` (71), `forkDaemon` → `forkDetach` (7), `catchAllCause` → `catchCause` (1). The two longer `catch*` names had to go first, or a bare `catchAll` match eats them.

**Removals needing a judgement call:** `zipRight` → `andThen` (v4's `andThen` takes a bare Effect in pipe position — same sequencing, confirmed against its signature), `dieMessage` → `die(new Error(…))`, `Cause.failureOption` → `Cause.findErrorOption`, `Layer.scoped` → `Layer.effect`.

> [!important] Red by design — do not enable auto-merge
> 728 errors remain, all Logger (#900) or Schema (#901, #902). Per #896 there is no green intermediate state; #902 is the first phase whose CI can pass.

## Workspaces affected

`@osn/api`, `@osn/client`, `@pulse/api`, `@zap/api`, `@cire/api`, `@shared/crypto`, `@shared/email`, `@shared/redis`. Two changesets, split so version-less `@cire/*` never shares one with a versioned package.

## Issues

- Closes #899
- Part of #895; follows #897/#898 (#906)

### `Either` → `Result` is more than an import swap

`Result`'s variants are tagged `Success`/`Failure`, not `Right`/`Left`, and carry `success`/`failure` rather than `right`/`left` — 71 accessor sites and 69 tag sites.

The tag comparisons written as **string literals are invisible to the compiler** (`toBe` takes `any`). That is the trap phase 0 found, and it was wider than the spike suggested: 9 files, including hand-rolled `{ _tag: "Left"; left: … }` union types in `osn/client/src/service.ts` that mirrored `Either` structurally. Those were found by grep, not by `tsc`. The accessor half I deliberately left for `tsc` to locate rather than blanket-renaming `.left`/`.right`, which would have hit unrelated objects.

**One file was deliberately left alone**: `cire/api/tests/schemas/vendors.test.ts` decodes through `Schema.decodeUnknownEither`, which maps to `decodeUnknownExit` — an `Exit`, not a `Result`. Its tags belong to the Schema phase. The three sibling schema tests *do* wrap in `Effect.result` and were converted.

### The `Runtime.isFiberFailure` removal changes behaviour on a security path

`Runtime.isFiberFailure` and `FiberFailureCauseId` are gone. v4's runner rejects with `Cause.squash(cause)` — verified in v4 source, where `causeSquash` returns `partitioned.Fail[0].error` directly — so `osn/api/src/lib/safe-error.ts` and `grant-failure.ts` have nothing left to unwrap and collapse to a narrowing check.

**That carries a real change, and both helpers now document it rather than hiding it.** `Cause.squash` surfaces a *defect* when the cause carries no `Fail`, where v3's `Cause.failureOption` returned `None` and the helpers answered `null`.

- In `safe-error.ts` the **allowlist contains it**: a defect can only leak a message if it is an `Error` carrying a `_tag` that is also allow-listed, and every allow-listed tag belongs to a `Data.TaggedError` raised through `Effect.fail` — which is a `Fail`, not a defect.
- In `grant-failure.ts` it moves the cookie marker in the **conservative** direction: a defect tagged `DatabaseError` now reads as "status unknown" and keeps the marker, where before it retracted. A kept marker costs a pointless grant; a wrongly-retracted one costs a sign-in.

Preserving v3 semantics exactly would need an `Exit`-returning runner threaded through `makeAppRunner` and every route handler. That is wider than this phase, and I did not do it. **Flagging it for a call:** if you want the exact old semantics, that is its own piece of work.

### Added the test that was missing under the riskiest change

`Layer.scoped` → `Layer.effect` keeps the finalizer only because v4's `Layer.effect` returns `Layer<I, E, Exclude<R, Scope>>` — it supplies the Scope, so `Effect.addFinalizer` still runs. A rewrite that dropped the scope would **type-check and leak Redis connections**, and since the limiters fail closed that surfaces as rejected requests rather than a crash.

The suite covered `client.quit()` delegation but **never asserted the layer's finalizer fires**. So the one change most able to fail silently had neither compiler nor test coverage. Added a test that builds the layer in a scope, closes it, and checks the memory client's store was cleared — `quit()` clears the store, which makes the finalizer observable. It passes, and that is the evidence.

### Two `{ cause: { error } }` casts removed

v4 flattened `Cause` into an array of reasons, so the hand-rolled casts in the email tests no longer describe the shape. They now narrow the `Exit` and read through `Cause.findErrorOption` — less casting than before, not more.

### `--no-verify`, again

Same reason phase 1 recorded: files this phase touched still contain Schema and Logger calls the pre-commit lint hook rejects. `fmt:check` was run separately and passes, so only the lint half was bypassed.

## Test plan

- **Error inventory: 1112 → 728**, and every remaining error is Logger or Schema. Confirmed no phase-2 symbol survives anywhere: `catchAll`, `either`, `zipRight`, `forkDaemon`, `dieMessage`, `Layer.scoped`, `Cause.failureOption`, `Runtime.isFiberFailure`, `Either.` all report **0** occurrences in tracked source.
- **Six packages fully green, 460 tests passing** — `@shared/db-utils` 59, `@shared/crypto` 84, `@shared/redis` 81, `@osn/db` 107, `@pulse/db` 104, `@zap/db` 26. `@shared/crypto` and `@shared/redis` are new this phase.
- **The new finalizer test passes** (`shared/redis` 80 → 81).
- **`bun run fmt:check`** passes.
- **`scripts/validate-changesets.sh`** passes.

Not run, deferred to #903: full suite, both D1 tiers, `bun run build`, dev-tier verification. Nothing deploys off `effect-v4`.

---


chav-aniket:
## CI is red on `708b1d5`, by design — and the failures moved in the right direction

Not fixing them here. These are this PR's failures, caused deliberately, and they close in the later phases of this stack. Per #896 there is no green intermediate state; #902 is the first phase whose CI can pass.

| Check | Cause | Fixed by |
| --- | --- | --- |
| Type Check | 728 errors, all Logger or Schema | #900, #901, #902 |
| Lint & Format | Same inventory via oxlint's `import(namespace)` rule. **`fmt:check` alone passes** | #900–#902 |
| OpenAPI Freshness | Runtime crash — see below | #901, #902 |

**Do not enable auto-merge.** It would wait on checks that cannot go green until the end of the stack.

### The failures shrank, and that is the signal worth reading

Phase 1 (#906) left **1112** errors. This PR leaves **728**. Every phase-2 symbol is gone from tracked source — `catchAll`, `catchAllDefect`, `catchAllCause`, `either`, `zipRight`, `forkDaemon`, `dieMessage`, `Layer.scoped`, `Cause.failureOption`, `Runtime.isFiberFailure`, `Either.` all report **0** occurrences. I checked the lint output for any symbol outside the Logger and Schema families and found none.

### OpenAPI Freshness moved to the next blocker

Worth noting because it confirms the phase ordering is doing what it should. On #906 this job crashed at:

```
TypeError: Layer.scoped is not a function.
    at shared/redis/src/service.ts:36:58
```

That was phase 2's work, and it is fixed. The same job now crashes at:

```
TypeError: Schema.minItems is not a function.
    at pulse/api/src/services/events.ts:62:10
```

Phase 4's work (#901). As #906's comment explained, this job *executes* the API rather than reading its types, so it clears only when the last runtime-breaking call goes — which is the Schema phases, not the type errors. The crash walking forward one phase at a time is the expected progression, and `git status shared/openapi/` stays clean, so no committed spec has drifted.

### Not re-running anything

Deterministic compile and import failures, not flakes. A re-run produces the identical result.


---

chav-aniket:
## `Build & Test` and `CI` — completing the picture from my earlier comment

Both failed on `708b1d5` after I posted above; that comment named only three checks because these two had not reported yet. Same by-design cause, but **the `Build & Test` log carried something new and worth having.**

### CI ran the tests, and they mostly pass

Unlike the type-only checks, this job got through `bun run build` and executed the suites: **1223 tests passed**, with nine packages failing — `@cire/api`, `@osn/api`, `@osn/client`, `@osn/social`, `@osn/ui`, `@pulse/api`, `@shared/email`, `@shared/observability`, `@zap/api`. Every one traces to a `Schema.*` or `Logger.*` call evaluated at import time, i.e. #900–#902.

Worth noting that a **local** `bun run build` cannot tell you this: it dies in a sandbox on `fonts.google.com` (`@cire/landing`), which stops turbo before the tests ever run. CI has network and gets through. A local `bun run test` reproduces the same nine packages once the build is out of the way.

### The finding: a package with zero Effect imports can still be broken

`@osn/ui` and `@osn/social` **import Effect nowhere**. Both depend on `@osn/client`, whose `src/tokens.ts:87` calls `Schema.Record({ key, value })` — v4 takes them as two positional arguments, so it throws while the module is still being evaluated:

```
TypeError: undefined is not an object (evaluating 'value.ast')
  at effect/src/Schema.ts:3965
  at osn/client/src/tokens.ts:87
```

Their suites cannot run at all until that one line is fixed: 10 of 19 test files in `@osn/social`, one in `@osn/ui`.

**Not fixing it here** — it is a `Schema` signature change, squarely #901's scope, and this PR should not widen into the next phase. But two things came out of it and are now recorded in `wiki/runbooks/effect-v4-migration.md` and on #901:

- **`osn/client/src/tokens.ts` is the highest-leverage file in the migration.** One signature fix unblocks three test suites, two of which are not otherwise part of this work. #901 now says to do it first.
- **A package's Effect surface is not the measure of its exposure.** The plan sized the Schema work by counting files that *import* `Schema` — 58, in six packages. It never counted packages that merely depend on those and fail at module load. Only a runtime signal finds those; a type-check of `@osn/ui` says nothing.

### Not re-running anything

Deterministic import-time failures, not flakes.


---

chav-aniket:
## 📚 Effect v4 stack — 2 of 6 ← **you are here**

| # | PR | Base | CI |
|---|---|---|---|
| **2** | **#908 — renames, `Cause`/`Runtime`** | #906 | 🔴 by design |

**Merge bottom-up into `effect-v4`**, then `effect-v4` → `main`. Do **not** enable auto-merge on 1–4; their CI cannot go green (#896). The tree goes green at #912.

**Follow-up on this PR's open question.** Its "Decisions" section flagged that `Cause.squash` surfaces a defect where v3's `Cause.failureOption` returned `None`, and left restoring the exact old semantics as a call to make. **That call was made and the work is done in #913**: `makeAppRunner` now goes through `runPromiseExit` and reads the `Fail` with `Cause.findErrorOption`, so the v3 contract is back for all 136 call sites with a diff in one file.

The containment argument written here — that the allowlist makes it safe — turned out to be **false**, and #913 has the reproduction: `Effect.die`, `Effect.orDie` (already used at three sites in this repo) and a bare `throw` inside `Effect.sync` all put an allow-listed tag in front of the allowlist.

```bash
```

---

chav-aniket:
**Landed.** `effect-v4` was fast-forwarded to the top of the stack (`804e8d4` → `5ab1418`, 20 commits), so this branch's commits are in `effect-v4` — verified with `git merge-base --is-ancestor`.

Closing rather than merging: GitHub does not auto-close a PR whose commits reach the base by direct push (that only happens for the default branch), and this PR's CI is red by design (#896), so there was no merge button to press. The fast-forward advances the integration branch without overriding branch protection.

**#899 is not closed by this** — closing keywords only bind on a PR targeting the default branch. The `effect-v4` → `main` PR carries `Closes #896`…`#903`.

The open question this PR raised — restoring exact v3 `Exit` semantics after `Cause.squash` — was answered in #913: the containment argument here turned out to be **false**, and the fix moved to `makeAppRunner` via `runPromiseExit` + `Cause.findErrorOption`.

---

PR #909, opened 2026-09-06
feat(effect): rework Logger for v4, and fix a secret leak in annotation redaction

### LINKED ISSUE #900: Effect v4 phase 3: rewrite Logger/LogLevel and the @effect/opentelemetry module renames

chav-aniket:
## Summary

Phase 3 of the Effect v4 migration (#895), closing #900. Stacked on #908 (phase 2).

**This phase found a live secret leak that predates the migration**, and fixing it is the larger half of the diff. See Decisions.

The mechanical half: v4 rewrote `Logger` and made `LogLevel` string literals. `Logger.replace(defaultLogger, x)` → `Logger.layer([x])`, `Logger.pretty` → a layer of `consolePretty()` + `tracerLogger`, `Logger.withMinimumLogLevel(LogLevel.X)` → `Effect.provideService(References.MinimumLogLevel, "X")`. Note **`"Warn"`, not v3's `"Warning"`**.

> [!important] Red by design — do not enable auto-merge
> 621 errors remain, **all** Schema (#901, #902). Zero Logger or LogLevel errors anywhere. #902 is the first phase whose CI can pass.

## Workspaces affected

`@shared/observability`, `@shared/email`, `@osn/api`, `@pulse/api`, `@zap/api`. One changeset; no `@cire/*` package is touched.

## Issues

- Closes #900
- Part of #895; follows #899 (#908)
- **A security finding here needs its own tracker issue** — see below. It is not filed yet because `xchromo/osn-tracker` is not in this session's scope.

### The annotation redaction has never worked — this fixes it

`redact()` matches the deny-list against an object's **keys**. The v3 logger mapped over each annotation **value**:

```ts
annotations: HashMap.map(options.annotations, (value) => redact(value))
```

So `redact` only ever saw a bare scalar with no key attached, and passed it through. **`Effect.annotateLogs({ accessToken })` reached the sink in clear** — as did `refreshToken`, `idToken`, `email`, `handle`, and every other deny-listed key, on every tier, for as long as this has been in place.

Verified rather than inferred: restoring the per-value form fails the new test with `expected 'eyJsecret' to be '[REDACTED]'`.

**This affects `main` today** and is fixable there independently of this migration — the same one-line change applies to the v3 code. It should not wait for the Schema phases.

**Nothing caught it because the test named for catching it didn't test it.** `layer.test.ts` had a case called *"end-to-end: `Effect.logInfo` with secret annotations emits a redacted entry"* which built the layer into an unused `_loggerLayer`, provided a **raw** capture logger with no redaction in the chain, and asserted the annotation came through **unredacted**. Its own comment admitted this; only the name was wrong. Redaction was covered by `redact.test.ts` at the pure-function level and nowhere else.

### Redaction moves to the output side, because v4 removed the input

v4 moved annotations off `Logger.Options` (now just `message`, `logLevel`, `cause`, `fiber`, `date`) and onto the fiber, read via `References.CurrentLogAnnotations`. There is no input to intercept any more, and **dropping the annotation pass compiles cleanly** while every annotated secret reaches the sink.

The replacement wraps `Logger.formatStructured`, whose output object already carries the resolved `message` and `annotations` record. Arguably better than the v3 shape: it scrubs whatever the formatter resolved, wherever it came from.

### `Logger.tracerLogger` gets one definition, not eleven

`Logger.layer` replaces the **whole** active set — unlike v3's `Logger.replace`, which swapped one logger and left the rest standing. Omit `tracerLogger` and log-to-span correlation disappears with no error and no failing type-check.

The four dev entrypoints provided `Logger.pretty` in eleven places. Rather than inlining `Logger.layer([consolePretty(), tracerLogger])` eleven times — one careless edit away from losing it in one of them — that is now `PrettyLoggerLive`, exported once from `@shared/observability`.

Before adding the import to zap's **Worker** entry I checked that `@shared/observability` is already in that bundle's graph via `app.ts`, so this pulls in nothing new and does not disturb the workerd carve-out.

`{ mergeWithExisting: true }` is **not** the v3 equivalent, incidentally: it *adds* to the set, leaving the default logger running alongside so every line is emitted twice.

### Local loses ANSI colour

`consolePretty` is opaque — no seam to redact through it. So local now gets an indented structured rendering rather than coloured pretty output. The trade is one redaction point covering every tier over prettier local logs. **Reversible if that is the wrong call**, at the cost of local logs going unredacted.

### The JSON severity field is now `level`

v4's structured output names it `level`, not `logLevel`. **Grafana queries, panels and alerts filtering on the old name match nothing.** No dashboards are committed to this repo — they live in Grafana Cloud, so that migration is yours to do by hand. Recorded in `wiki/observability/logging.md`.

### `wiki/observability/logging.md` overstated the control

Its §Redaction claimed coverage the code did not deliver. It now records what actually happens, names the v3 bug rather than editing it out, and carries the `level` rename.

## Test plan

- **A real end-to-end redaction test**, replacing the misnamed one. Runs the actual layer, reads what reaches stdout, asserts a deny-listed key is `[REDACTED]` and an allow-listed one survives. **Verified non-vacuous**: restoring the per-value form fails it.
- **621 errors, down from 728**, and **zero** are Logger or LogLevel — the signal this phase is complete. Every remaining one is Schema.
- **608 tests pass across 8 fully-green packages**: `@shared/observability` 92, `@shared/email` 55, `@shared/db-utils` 59, `@shared/crypto` 84, `@shared/redis` 81, `@osn/db` 107, `@pulse/db` 104, `@zap/db` 26. Observability and email are new this phase.
- The existing `makeLoggerLayer output format` block still passes — it runs the real layer per tier and treats empty output as failure, which is the "deployed tier goes silent" guard.
- **`bun run fmt:check`** passes; **`scripts/validate-changesets.sh`** passes.

Not run, deferred to #903: full suite, both D1 tiers, `bun run build`, dev-tier verification. Nothing deploys off `effect-v4`.

---

### DIFF
--- a/.changeset/effect-v4-phase-3.md
+++ b/.changeset/effect-v4-phase-3.md
@@ -1,0 +1,34 @@
+---
+"@shared/observability": minor
+"@shared/email": patch
+"@osn/api": patch
+"@pulse/api": patch
+"@zap/api": patch
+---
+
+Rebuild the logger for Effect v4, and fix a secret leak in annotation redaction.
+
+`redact()` matches the deny-list against an object's **keys**, and the v3 logger
+mapped over each annotation **value** — so it only ever saw a bare scalar with no
+key attached and passed it through. `Effect.annotateLogs({ accessToken })`
+reached the sink in clear, along with every other deny-listed key, on every tier.
+The record is now passed whole.
+
+v4 moved annotations off the logger's `Options` and onto the fiber, so redaction
+moves to the output side, wrapping `Logger.formatStructured`. `Logger.layer`
+replaces the whole active set, so `Logger.tracerLogger` is listed explicitly —
+omitting it drops log-to-span correlation silently. `LogLevel` is now string
+literals (`"Warn"`, not v3's `"Warning"`), and the minimum level is a
+`References.MinimumLogLevel` service rather than `Logger.minimumLogLevel`.
+
+Adds `PrettyLoggerLive` for the dev-server entrypoints, replacing v3's
+`Logger.pretty`. It exists as one export rather than eleven inline
+`Logger.layer([…])` arrays so `tracerLogger` has a single place to be got right.
+
+Local output loses ANSI colour for an indented structured rendering:
+`consolePretty` is opaque, so there is no seam to redact through it, and one
+redaction point covering every tier is the better trade.
+
+**The JSON severity field is now `level`, not `logLevel`.** Grafana queries,
+panels and alerts filtering on the old name match nothing and must be updated in
+Grafana Cloud by hand.
--- a/osn/api/src/local.ts
+++ b/osn/api/src/local.ts
@@ -1,5 +1,5 @@
 import { DbLive } from "@osn/db/service";
-import { initObservability } from "@shared/observability";
+import { initObservability, PrettyLoggerLive } from "@shared/observability";
 import { Effect, Layer, Logger } from "effect";
 
 import { createApp, type App } from "./app";
@@ -107,7 +107,7 @@ export function startBunServer(
       yield* Effect.logInfo("osn-app listening");
     }).pipe(
       Effect.annotateLogs({ port: String(port), service: SERVICE_NAME }),
-      Effect.provide(Logger.pretty),
+      Effect.provide(PrettyLoggerLive),
       Effect.provide(observabilityLayer),
     ),
   );
--- a/osn/api/tests/lib/email-layer.test.ts
+++ b/osn/api/tests/lib/email-layer.test.ts
@@ -62,13 +62,13 @@ describe("selectEmailLayer", () => {
 
   it("creds absent + non-local + opt-in set → emits a loud degraded-mode startup warning (no PII)", () => {
     const lines: string[] = [];
-    const captureLayer = Logger.replace(
-      Logger.defaultLogger,
+    const captureLayer = Logger.layer([
+      // v4 log levels are string literals, so there is no `.label` to read.
       Logger.make(({ message, logLevel }) => {
         const msg = Array.isArray(message) ? message.join(" ") : String(message);
-        lines.push(`[${logLevel.label}] ${msg}`);
+        lines.push(`[${logLevel}] ${msg}`);
       }),
-    );
+    ]);
 
     // Selection emits the loud warning synchronously through the supplied
     // observability layer.
--- a/osn/api/tests/services/auth.test.ts
+++ b/osn/api/tests/services/auth.test.ts
@@ -2,7 +2,7 @@ import { it, expect, describe } from "@effect/vitest";
 import { passkeys } from "@osn/db/schema";
 import { Db } from "@osn/db/service";
 import { EmailError, EmailService, makeLogEmailLive } from "@shared/email";
-import { Effect, Layer, Logger, LogLevel } from "effect";
+import { Effect, Layer, Logger, References } from "effect";
 import { beforeAll } from "vitest";
 
 import { createInMemoryRotatedSessionStore } from "../../src/lib/rotated-session-store";
@@ -937,7 +937,7 @@ describe("LogEmailLive local-mode behaviour", () => {
         : String(options.message);
       captured.push(msg);
     });
-    const loggerLayer = Logger.replace(Logger.defaultLogger, sinkLogger);
+    const loggerLayer = Logger.layer([sinkLogger]);
     return { captured, loggerLayer };
   }
 
@@ -947,7 +947,10 @@ describe("LogEmailLive local-mode behaviour", () => {
     return Effect.gen(function* () {
       yield* svc
         .beginRegistration("dev@example.com", "devuser", "1990-01-01", "Dev User")
-        .pipe(Effect.provide(loggerLayer), Logger.withMinimumLogLevel(LogLevel.Debug));
+        .pipe(
+          Effect.provide(loggerLayer),
+          Effect.provideService(References.MinimumLogLevel, "Debug"),
+        );
 
       // Log line: template + subject + to, no OTP code.
       const emailLogs = logLines.filter((l) => l.includes("[email:log]"));
--- a/osn/api/tests/services/auth/helpers.test.ts
+++ b/osn/api/tests/services/auth/helpers.test.ts
@@ -1,4 +1,4 @@
-import { Effect, Logger, LogLevel } from "effect";
+import { Effect, Logger, References } from "effect";
 import { afterEach, describe, expect, it, vi } from "vitest";
 
 /**
@@ -38,14 +38,13 @@ async function capture(effect: Effect.Effect<void>): Promise<string[]> {
   await Effect.runPromise(
     effect.pipe(
       Effect.provide(
-        Logger.replace(
-          Logger.defaultLogger,
+        Logger.layer([
           Logger.make(({ message }) => {
             lines.push(String(message));
           }),
-        ),
+        ]),
       ),
-      Logger.withMinimumLogLevel(LogLevel.Debug),
+      Effect.provideService(References.MinimumLogLevel, "Debug"),
     ),
   );
   return lines;
--- a/pulse/api/src/local.ts
+++ b/pulse/api/src/local.ts
@@ -1,5 +1,5 @@
 import { DbLive } from "@pulse/db/service";
-import { initObservability } from "@shared/observability";
+import { initObservability, PrettyLoggerLive } from "@shared/observability";
 import type { ClientIpOptions } from "@shared/rate-limit";
 import { Effect, Logger } from "effect";
 
@@ -94,7 +94,7 @@ if (nonLocal && !oidc) {
   void Effect.runPromise(
     Effect.logError(
       "pulse-api: OIDC sign-in disabled — set OSN_ISSUER_URL, OSN_JWKS_URL, PULSE_API_ORIGIN, OSN_OIDC_CLIENT_ID and OSN_OIDC_CLIENT_SECRET to enable it",
-    ).pipe(Effect.provide(Logger.pretty), Effect.provide(observabilityLayer)),
+    ).pipe(Effect.provide(PrettyLoggerLive), Effect.provide(observabilityLayer)),
   ).catch(() => undefined);
 }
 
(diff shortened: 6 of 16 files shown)


chav-aniket:
## CI red on `466f81c`, by design — and every remaining failure is Schema

Not fixing them here. These are this PR's failures, caused deliberately, and they close in #901 and #902. Per #896 there is no green intermediate state; #902 is the first phase whose CI can pass.

| Check | Cause | Fixed by |
| --- | --- | --- |
| Lint & Format | 621 errors via oxlint's `import(namespace)` rule, all `Schema.*`. **`fmt:check` alone passes** | #901, #902 |
| OpenAPI Freshness | Runtime crash — see below | #901, #902 |
| Type Check | Same 621 | #901, #902 |
| Build & Test | Same, plus the nine packages whose suites fail at import time | #901, #902 |
| Require changeset | ✅ success | — |
| Scripts | ✅ success | — |

**Do not enable auto-merge.**

### The Logger work is complete, and the numbers say so

728 errors after phase 2 → **621** here, and **zero** are Logger or LogLevel. I checked the whole inventory for any symbol outside the Schema family and found none. Every one of the eight fully-green packages passes its suite — **608 tests** — with `@shared/observability` (92) and `@shared/email` (55) new this phase.

### OpenAPI Freshness moved on again

The useful signal from this job, third phase running. It *executes* the API rather than reading its types, so it clears only when the last runtime-breaking call goes:

| Phase | Crash |
| --- | --- |
| #906 | `TypeError: Layer.scoped is not a function` — `shared/redis/src/service.ts:36` |
| #908 | `TypeError: Schema.minItems is not a function` — `pulse/api/src/services/events.ts:62` |
| #909 (here) | `pulse/api` still `Schema.minItems`; `osn/api` now `TypeError: Schema.filter is not a function` — `osn/api/src/services/auth/helpers.ts:328` |

Both generators are now blocked purely on `Schema`, which is #901's work. `git status shared/openapi/` is clean, so no committed spec has drifted behind the crash.

### One thing on this PR is not a migration matter

The annotation redaction bug this PR fixes **exists on `main` right now** — `Effect.annotateLogs({ accessToken })` logs the token in clear, and has for as long as that code has existed. The fix here is stacked four PRs deep and will not reach `main` for a while.

The same one-line change applies to the v3 code unchanged. It should not wait on this stack, and it wants a tracker issue rather than a public one. Both are with the repo owner.

### Not re-running anything

Deterministic compile and import failures, not flakes.


---

chav-aniket:
Correction to the table above: `Build & Test` reports **seven** failing packages on this head, not nine.

```
Failed: @cire/api#test, @osn/api#test, @osn/client#test,
        @osn/social#test, @osn/ui#test, @pulse/api#test, @zap/api#test
```

I carried the nine over from #908 while the job was still in progress. The two that dropped off are `@shared/email` and `@shared/observability` — precisely the ones this phase finishes — so the number moving is the evidence that the Logger work landed, and it was worth not leaving it stale.

The remaining seven are unchanged in kind: all fail at import time on a `Schema.*` call, and all close in #901 and #902.


---

chav-aniket:
## Correction: I overstated the redaction bug

My comment above, and this PR's body, describe the annotation-redaction bug as an active leak — *"`Effect.annotateLogs({ accessToken })` reached the sink in clear"*. **That was wrong**, and I only found the bound after writing it. Two things limit it, both now verified:

- **Only top-level annotation keys escaped.** `redact(value)` still ran on every value, so a secret nested inside an annotation object was scrubbed exactly as intended.
- **No call site annotates a deny-listed key.** The annotation keys actually in use across the repo are `service`, `reason`, `error`, `port`, `entitlement`, `enquiryId`, `template`, `tag`, `limit`, `directoryVendorId`, `declared`, `contentType`. None is on the deny-list.

So **nothing has been leaking**. What was broken is the control, not the output: the first `Effect.annotateLogs({ accessToken })` anyone wrote would have gone out in clear, silently, with no failing test to catch it. Latent, not active — and rated medium rather than high on that basis.

Everything else in the analysis stands: the mechanism was wrong, the guard test was fake, and both are fixed.

**The `main` fix is now open as #910**, independent of this stack, with the finding recorded privately in `xchromo/osn-tracker#637`. The v4 shape here (redacting the resolved record on the output side) is unchanged and still correct — it was never the per-value form.


---

chav-aniket:
## 📚 Effect v4 stack — 3 of 6 ← **you are here**

| # | PR | Base | CI |
|---|---|---|---|
| **3** | **#909 — Logger rework** | #908 | 🔴 by design |

**Merge bottom-up into `effect-v4`**, then `effect-v4` → `main`. Do **not** enable auto-merge on 1–4; their CI cannot go green (#896).

### Two claims in this PR's description are now known to be wrong — both corrected in #913

**1. "`consolePretty` is opaque — no seam to redact through it."** False. v4 exposes the entry on the *input* side: `Logger.Options` carries `message`, and the pretty logger reads annotations via `fiber.getRef(References.CurrentLogAnnotations)`. Shadowing both and delegating to an untouched `consolePretty` redacts it while Effect keeps colour, spans, `LogToStderr` and the fiber id.

So the trade recorded here — *"local loses ANSI colour… one redaction point covering every tier over prettier local logs"* — **was never a real trade**. #913 restores colour to `local` *and* redacts it, and makes `PrettyLoggerLive` redacting too, which removes the unredacted-logger category from the codebase entirely.

**2. The description called the annotation redaction bug a live leak.** It is not: no call site annotates a deny-listed key. That was corrected on #910 at the time; repeating it here so this description is not read at face value later.

### Also traceable to this PR

`zap/api/src/index.ts` — a **deployed** Worker — was left on the unredacted `PrettyLoggerLive` by this phase's mechanical `Logger.pretty` → `PrettyLoggerLive` rename. Pre-existing (v3's `Logger.pretty` was equally unredacted at the same two lines), fixed in #913. Filed alongside it: `xchromo/osn-tracker#639`, that `Logger.tracerLogger` ships raw annotations, messages and causes to Grafana on **every** tier, bypassing the deny-list — which this PR's `Logger.layer([redacting, tracerLogger])` shape does not address.

```bash
gh stack link 906 908 909 911 912 913
```

---

chav-aniket:

Closing rather than merging: GitHub does not auto-close on a direct push to a non-default base, and this PR's CI is red by design (#896). **#900 is not closed by this** — the `effect-v4` → `main` PR carries `Closes #896`…`#903`.

Two claims in this PR's description were later shown wrong and are corrected in #913: `Logger.consolePretty()` is **not** opaque (so `local` losing ANSI colour was an unnecessary trade — it is now redacted *and* coloured), and the annotation-redaction bug was latent rather than a live leak.

---

PR #911, opened 2026-09-06
feat(effect): migrate the Schema surface of osn-api, pulse-api, zap-api and the client

### LINKED ISSUE #901: Effect v4 phase 4: migrate Schema in osn, pulse, zap, cire/theme and osn/client (20 files)

chav-aniket:
Phase 4 of the Effect v4 migration (#901), stacked on #909.

After this, **36 of 37 packages type-check**. The only one left is `@cire/api`, which is phase 5 (#902).

## Issues

- Closes #901 (Schema in osn/api, pulse/api, zap/api, cire/theme, osn/client)
- Part of #895; follows #900 (#909)

> [!note] This keyword will not auto-close #901 from here
> GitHub only creates the linked-PR relationship for a pull request targeting the repository's **default branch**. This one targets `claude/effect-v4-phase-3`, so the keyword records intent and nothing more — which is why the sub-issues still show as open with no linked PR. The `effect-v4` → `main` PR at the top of the stack is what has to carry the full `Closes #896`…`#903` set, and it is what will actually close them.

### Checks replace the constraint combinators

v3's constraint combinators are v4 *checks*, applied through a schema's `.check(...)` rather than `.pipe(...)`:

| v3 | v4 |
|---|---|
| `Schema.maxLength(n)` / `minLength(n)` | `Schema.isMaxLength(n)` / `isMinLength(n)` |
| `Schema.maxItems(n)` / `minItems(n)` | the same two — `isMinLength` doubles as the collection-size check |
| `Schema.int()` | `Schema.isInt()` |
| `Schema.between(a, b)` | `Schema.isBetween({ minimum: a, maximum: b })` |
| `Schema.filter(pred)` | `Schema.check(Schema.makeFilter(pred))` |
| `Schema.Literal(a, b, …)` | `Schema.Literals([a, b, …])` |
| `Schema.decodeUnknown` | `Schema.decodeUnknownEffect` |
| `Schema.Record({ key, value })` | `Schema.Record(key, value)` |

`isBetween` is inclusive at both ends, same as v3's `between`, so **no range moved**.

A v4 filter carries its own failure message in its return value — `undefined`/`true` is success, a string is the message — so `{ message: () => "…" }` becomes the predicate returning that string. **Every validator keeps its exact wording.** `BirthdateSchema`'s multi-branch predicate returns the message from each rejecting branch rather than `false`. Where a `message` annotation genuinely belongs on the check itself (`PriceAmountMajor`), v4 takes it as a plain string rather than a thunk.

### Three copies of a workaround are deleted, not ported

`@pulse/api`'s events, series and discovery services each carried a hand-rolled "validate the string, then transform to a Date" pair. The comment in `events.ts` said why:

> `Schema.DateFromString` in this Effect version allows Invalid Date — use a validated transform

v4's rejects it. Verified directly against the resolved package: `"not-a-date"` and `""` both fail with `Expected a valid Date`, a valid string still decodes, and encoding still yields an ISO string. So all three collapse back to the stock `Schema.DateFromString`, and the `Schema.transform` → `decodeTo`/`SchemaTransformation` migration is not needed here at all.

### `isAuthExpiredError` and the missing wrapper

v4 removed `FiberFailure`. `runPromise` now rejects with `Cause.squash(cause)`, which for a typed failure is the error instance itself — so `escaped instanceof AuthExpiredError` is `true` where the test asserted `false`.

The predicate keeps all three arms, but the comments no longer claim the wrapper is what arrives: `instanceof` now carries the common path, and the printout arm stays for a consumer bundle built against v3 (`cire/host`, `cire/vendor`) and for any boundary that strips both the prototype and the `_tag`. The escape test pins the v4 shape, with an explicit v3-printout case beside it so that arm keeps its coverage.

### Test assertions that were pinned to v3 output

- `osn/client/tests/tokens.test.ts` — v4 decode errors render a reason line plus an `at [path]` line, and no longer echo the offending value (v3 wrote `actual "not_a_number"`). All five assertions still pin **both halves** — which key, and why — so a schema change that starts rejecting the wrong field still fails here.
- `osn/api/tests/lib/email-layer.test.ts` — v4 log levels are capitalised string literals, so the rendered label is `[Warn]`, not `[WARN]`. Respelled the alternation rather than making it case-insensitive; a loose match would also accept a level that does not exist in v4's set.

`cire/theme` was on the phase-4 list but imports Effect nowhere — only three comments mention Schema. Nothing to migrate; it type-checks clean as-is.

## Verification

Type-checking a check proves nothing about whether it still *rejects*, so each was exercised against its package's own resolved `effect`:

| Package | Checked |
|---|---|
| `@zap/api` | over-long body, empty body, too few / too many array items |
| `@pulse/api` | duplicate `commsChannels`, empty channels, unknown channel, each arm of the discovery composite filter (partial location triangle, price without currency, min > max, half a cursor), lat/limit ranges, invalid date |
| `@osn/api` | address with no `@`, address over the 255-char RFC 5321 ceiling, uppercase / over-length / empty handle, `2021-02-30`, a future birthdate, a non-zero-padded date, blank / over-64-char passkey label — each with its original message |

Suites:

| Package | Result |
|---|---|
| `@osn/api` | 1141 passed |
| `@pulse/api` | 647 passed |
| `@zap/api` | 179 passed |
| `@osn/client` | 228 passed |
| `@osn/ui` | 143 passed |
| `@osn/social` | 147 passed |

`@osn/ui` and `@osn/social` import Effect nowhere and were red only because they load it through `@osn/client`; the one-line `Schema.Record` fix unblocked both, as #901 predicted.

## Still red by design

`@cire/api` — phase 5 (#902). CI on this branch will fail on it, as it did on #906, #908 and #909. The tree goes green when phase 5 lands.

---

### DIFF
--- a/.changeset/effect-v4-phase-4.md
+++ b/.changeset/effect-v4-phase-4.md
@@ -1,0 +1,36 @@
+---
+"@osn/api": patch
+"@osn/client": patch
+"@pulse/api": patch
+"@zap/api": patch
+---
+
+Migrate the Effect Schema surface of the four service packages to v4.
+
+v3's constraint combinators are v4 *checks*, applied through a schema's
+`.check(...)` rather than `.pipe(...)`: `maxLength`/`minLength`/`minItems`/
+`maxItems` collapse onto `isMaxLength`/`isMinLength`, `int` becomes `isInt`, and
+`between(a, b)` becomes `isBetween({ minimum, maximum })` — still inclusive at
+both ends, so no range moved. `Schema.filter` becomes
+`Schema.check(Schema.makeFilter(…))`, and because a v4 filter carries its own
+failure message in its return value, the `{ message: () => "…" }` option becomes
+the predicate returning that string; every validator keeps its exact wording.
+`Schema.Literal` takes a single literal, so enums (and the spreads over
+`SUPPORTED_CURRENCIES`, `SHARE_SOURCES` and `INTEREST_CATEGORIES`) become
+`Schema.Literals([…])`. `Schema.decodeUnknown` becomes
+`Schema.decodeUnknownEffect`, and `Schema.Record` takes its key and value
+positionally.
+
+Three copies of a workaround are deleted rather than ported. `@pulse/api`'s
+events, series and discovery services each carried a hand-rolled "validate the
+string, then transform to a Date" pair because v3's `DateFromString` accepted a
+string that parses to an Invalid Date. v4's rejects it, so all three are now
+`Schema.DateFromString`.
+
+`@osn/client`'s `isAuthExpiredError` keeps all three of its arms, but the
+comments no longer claim a `FiberFailure` is what arrives: v4 removed the
+wrapper and `runPromise` rejects with the squashed error itself, so `instanceof`
+now carries the common path. The printout arm stays for a consumer bundle built
+against v3, and for any boundary that strips both the prototype and the `_tag`.
+
+Every migrated check was verified to still *reject*, not merely type-check.
--- a/osn/api/src/services/auth/email-change.ts
+++ b/osn/api/src/services/auth/email-change.ts
@@ -49,7 +49,7 @@ export function createEmailChangeModule(ctx: AuthContext, stepUp: StepUpModule) 
     Db | EmailService
   > =>
     Effect.gen(function* () {
-      yield* Schema.decodeUnknown(EmailSchema)(newEmail).pipe(
+      yield* Schema.decodeUnknownEffect(EmailSchema)(newEmail).pipe(
         Effect.mapError((cause) => new ValidationError({ cause })),
       );
       const normalised = newEmail.toLowerCase();
--- a/osn/api/src/services/auth/helpers.ts
+++ b/osn/api/src/services/auth/helpers.ts
@@ -324,16 +324,18 @@ export function sessionHandleFromHash(sessionHash: string): string {
 // Copenhagen Book M3: cap length at 255 (the practical RFC 5321 mailbox
 // ceiling) BEFORE the regex runs — rejects absurd payloads outright and
 // keeps the stored `accounts.email` column bounded.
-export const EmailSchema = Schema.String.pipe(
-  Schema.filter((s) => s.length <= 255 && /^[^\s@]+@[^\s@]+\.[^\s@]+$/.test(s), {
-    message: () => "Invalid email",
-  }),
+export const EmailSchema = Schema.String.check(
+  Schema.makeFilter((s) =>
+    s.length <= 255 && /^[^\s@]+@[^\s@]+\.[^\s@]+$/.test(s) ? undefined : "Invalid email",
+  ),
 );
 
-export const HandleSchema = Schema.String.pipe(
-  Schema.filter((s) => /^[a-z0-9_]{1,30}$/.test(s), {
-    message: () => "Handle must be 1–30 characters: lowercase letters, numbers, underscores only",
-  }),
+export const HandleSchema = Schema.String.check(
+  Schema.makeFilter((s) =>
+    /^[a-z0-9_]{1,30}$/.test(s)
+      ? undefined
+      : "Handle must be 1–30 characters: lowercase letters, numbers, underscores only",
+  ),
 );
 
 /**
@@ -344,19 +346,17 @@ export const HandleSchema = Schema.String.pipe(
  * COPPA-specific 422 rather than a generic 400. The value is never persisted.
  * See [[compliance/coppa]].
  */
-export const BirthdateSchema = Schema.String.pipe(
-  Schema.filter(
-    (s) => {
-      if (!/^\d{4}-\d{2}-\d{2}$/.test(s)) return false;
-      const d = new Date(`${s}T00:00:00.000Z`);
-      if (Number.isNaN(d.getTime())) return false;
-      // Reject dates that don't round-trip (e.g. 2021-02-30 → 2021-03-02).
-      if (d.toISOString().slice(0, 10) !== s) return false;
-      // A birthdate in the future is nonsensical.
-      return d.getTime() <= Date.now();
-    },
-    { message: () => "Invalid birthdate" },
-  ),
+export const BirthdateSchema = Schema.String.check(
+  Schema.makeFilter((s) => {
+    const invalid = "Invalid birthdate";
+    if (!/^\d{4}-\d{2}-\d{2}$/.test(s)) return invalid;
+    const d = new Date(`${s}T00:00:00.000Z`);
+    if (Number.isNaN(d.getTime())) return invalid;
+    // Reject dates that don't round-trip (e.g. 2021-02-30 → 2021-03-02).
+    if (d.toISOString().slice(0, 10) !== s) return invalid;
+    // A birthdate in the future is nonsensical.
+    return d.getTime() <= Date.now() ? undefined : invalid;
+  }),
 );
 
 /**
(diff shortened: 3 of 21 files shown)


chav-aniket:
CI is red here by design, as on #906, #908 and #909. Recording exactly what fails and why I am not fixing it *in this PR*.

**Failing:** Lint & Format, Type Check, Build & Test, and the CI rollup — all on `@cire/api`, which is still on Effect v3 at this point in the stack.

I checked the Lint & Format log rather than assuming, because that check was not in the set the earlier phases documented. It is the same root cause surfacing through a different tool — oxlint's `import` plugin resolves the identifiers against the v4 `Schema` namespace this branch now installs:

```
cire/api/tests/schemas/vendors.test.ts:14:60: "decodeUnknownEither" not found in imported namespace "./Schema.js"
cire/api/tests/schemas/budget.test.ts:16:39:  "decodeUnknown" not found in imported namespace "./Schema.js"
cire/api/tests/schemas/tasks.test.ts:9:39:    "decodeUnknown" not found in imported namespace "./Schema.js"
```

163 errors, every one of that shape. Nothing is failing for a reason outside the migration.

**The fix exists and is already pushed — as #912**, the next PR in this stack, which moves `@cire/api` to v4. It is green there: 9/9 checks, `bun run check` 37/37 packages, `bun run test` 38/38 tasks.

I have deliberately **not** ported it back into this PR. The usual rule is to port a known fix rather than wait on another PR to merge, but here the split *is* the deliverable: this is a stacked series on the long-lived `effect-v4` integration branch (the merge strategy chosen for #895), and Effect cannot be bumped one package at a time — its types cross every workspace boundary, which is what the phase-0 spike established. Folding #912 in would collapse two reviewable units into one ~60-file diff and leave the stack with no intermediate to review. The tree is red from #906 until #912 lands, and that was the plan going in.

No re-run spent — a re-run cannot change a deterministic type error.

---

chav-aniket:
## 📚 Effect v4 stack — 4 of 6 ← **you are here**

| # | PR | Base | CI |
|---|---|---|---|
| **4** | **#911 — Schema: osn/pulse/zap/client** | #909 | 🔴 by design |

**Merge bottom-up into `effect-v4`**, then `effect-v4` → `main`. This is the last red one: the tree goes green at #912.

The failure recorded in the standing-down comment above — 163 oxlint `import` errors resolving `Schema.decodeUnknown`/`decodeUnknownEither` against the v4 namespace — is fixed by #912, which moves `@cire/api` off v3. Nothing here needs another look.

```bash
```

---

chav-aniket:

Closing rather than merging: GitHub does not auto-close on a direct push to a non-default base, and this PR's CI is red by design (#896) — it was the last red one. **#901 is not closed by this**; the `effect-v4` → `main` PR carries `Closes #896`…`#903`.

---

PR #912, opened 2026-09-06
feat(effect): migrate @cire/api to v4 — the monorepo is green on Effect 4

### LINKED ISSUE #902: Effect v4 phase 5: migrate Schema in cire/api (38 files) — first green point

chav-aniket:
Phase 5 of the Effect v4 migration (#902), stacked on #911. `@cire/api` was the last package on v3.

**With this branch, the whole monorepo type-checks (37/37 packages) and passes its tests (38/38 tasks) under `effect@4.0.0-rc.112`.** This is the first green state since #906 — the tree has been red by design since phase 1.

## Issues

- Closes #902 (Schema in cire/api's 38 files)
- Part of #895; follows #901 (#911)

> [!note] This keyword will not auto-close #902 from here
> GitHub only creates the linked-PR relationship for a pull request targeting the repository's **default branch**. This one targets `claude/effect-v4-phase-4`, so the keyword records intent and nothing more. The `effect-v4` → `main` PR at the top of the stack is what has to carry the full `Closes #896`…`#903` set.

### `Schema.Union` takes one array

```diff
-Schema.Union(Schema.String, Schema.Null)
+Schema.Union([Schema.String, Schema.Null])
```

This is the quiet one. It does not error at the call site — the second member is simply dropped, and the failure surfaces somewhere downstream as `{} | null` where a `string | null` was expected. Worth knowing about because a union that silently loses a member is a validator that silently stops rejecting.

`Schema.Literal` behaves the same way (one literal in v4, several via `Schema.Literals([…])`), and it bit here in two forms the phase-4 regex didn't cover: `Schema.Literal("both", "events", "guests")` narrowed to `"both"`, and `Schema.Literal(...SERVICE_CATEGORIES.map((c) => c.key))` — a spread of a *computed* array, not an identifier.

### `optionalWith` with a default

22 sites, all the same shape:

```diff
-quantityWanted: Schema.optionalWith(Quantity, { default: () => 1 })
+quantityWanted: Quantity.pipe(Schema.withDecodingDefaultType(Effect.succeed(1)))
```

`withDecodingDefaultType` is the non-`exact` variant, so — as in v3 without `exact: true` — the default also applies when the key is present but `undefined`.

### The `ParseError` gap is answerable

Upstream's generated reference leaves `ParseResult.ParseError` as **"TODO: needs guidance"**, which is why phase 2 deferred it. The compiler answers it: v4 tags a decode failure `SchemaError`.

```diff
-Effect.catchTag("ParseError", () => badRequest(set))
+Effect.catchTag("SchemaError", () => badRequest(set))
```

41 `catchTag`/`catchTags` keys, plus one type annotation in `services/changes.ts` (`ParseResult.ParseError` → `Schema.SchemaError`).

### `Cause` is a list of reasons

```diff
-res.cause._tag === "Fail" && res.cause.error instanceof VendorNotInWedding
+Option.getOrUndefined(Cause.findErrorOption(res.cause)) instanceof VendorNotInWedding
```

`findErrorOption` returns `None` for a defect, so `undefined instanceof X` is `false` — the new form refuses a defect exactly as the `Fail` check did. This is the idiom `link-preview.test.ts` and `import.identity.test.ts` already use.

## Three workarounds deleted, and the guarantees they had

`Schema.transform` → `Schema.decodeTo(target, SchemaTransformation.transform(…))` was needed only three times: the time-zone canonicaliser, the HTTPS-URL normaliser, and the slider clamp. Three *other* uses were "trim, then bound the length", which collapse to a stock schema:

```diff
-Schema.String.pipe(
-  Schema.transform(Schema.String, { strict: true, decode: (s) => s.trim(), encode: (s) => s }),
-  Schema.minLength(1),
-  Schema.maxLength(64),
-)
+Schema.Trim.check(Schema.isMinLength(1), Schema.isMaxLength(64))
```

Two properties had to hold for that to be honest, and both were checked against the resolved package rather than assumed:

- **The checks run on the trimmed value.** `"   "` is rejected as blank, `"  abcde  "` passes a 5-char cap. If the checks ran on the input, both would be wrong in opposite directions.
- **`.check(a, b)` short-circuits on the first failure.** `settings.ts` documents why this matters: `TimeZone`'s `maxLength` runs before the ICU lookup precisely so an oversized blob never reaches it. v4 short-circuits — verified by counting calls into the second check after the first fails: zero.

## Two test fixes that are phase 3's, not Schema's

- `cire/api/tests/app.test.ts` isolated its `onError` log entry with `startsWith('{"')`. The local renderer is indented now, so that filter caught the string `"unhandled request error",` and nothing else. `Logger.layer` also **replaces** the whole active set, so the default logger whose `Fiber terminated…` stack dump the filter existed to exclude no longer exists — the whole capture *is* that one entry. Asserting on the capture makes the "no raw SQLite message leaked" check strictly stronger, since nothing is filtered out of it any more.
- Five `Logger` imports left dead by phase 3's move to `PrettyLoggerLive` (`osn/api`, `pulse/api`, `zap/api` × 2, `shared/observability` tests).

## Verification

| | |
|---|---|
| `bun run check` | **37 / 37 packages** |
| `bun run test` | **38 / 38 tasks** |
| `bun run lint` | clean (warnings only, none new) |
| `bun run fmt:check` | clean |
| `@cire/api` | 1752 passed, 0 failed |

No `^3.22` pin is left anywhere: 7 packages on `effect@4.0.0-rc.112`, 11 on `@effect/vitest@4.0.0-rc.112`.

## What is left

Phase 6 (#903): merge the stack into `effect-v4`, then `effect-v4` → `main`, dev-tier smoke, and the teardown.

---

### DIFF
--- a/.changeset/effect-v4-phase-5-cire.md
+++ b/.changeset/effect-v4-phase-5-cire.md
@@ -1,0 +1,40 @@
+---
+"@cire/api": patch
+---
+
+Migrate `@cire/api`'s Effect Schema surface to v4 — the last package on v3.
+**The whole monorepo now type-checks and passes its tests under Effect v4.**
+
+Everything phase 4 established applies here too (checks via `.check(...)`,
+`makeFilter` carrying its own message, `Literals([…])`, `decodeUnknownEffect`),
+plus four things cire hit first:
+
+- **`Schema.Union` takes one array**, not variadic members. Silent when it slips
+  through — a two-member union read as a single member, and the inference
+  degrades to `{}` rather than erroring at the call.
+- **`Schema.optionalWith(S, { default: () => v })` becomes
+  `S.pipe(Schema.withDecodingDefaultType(Effect.succeed(v)))`** — 22 sites, all
+  the same shape.
+- **A decode failure is tagged `SchemaError`, not `ParseError`.** Upstream's
+  generated reference leaves `ParseResult.ParseError` as "TODO: needs
+  guidance"; the compiler answers it. 41 `catchTag`/`catchTags` sites and one
+  type annotation in `services/changes.ts`.
+- **`Cause` is a list of reasons**, so `cause._tag === "Fail" && cause.error
+  instanceof X` becomes `Option.getOrUndefined(Cause.findErrorOption(cause))
+  instanceof X`. That still refuses a defect, exactly as the `Fail` check did.
+
+`Schema.transform` → `Schema.decodeTo(target, SchemaTransformation.transform(…))`
+was needed only twice (the time-zone canonicaliser and the URL normaliser) and
+once for a clamp. Three "trim, then bound the length" pairs collapse to
+`Schema.Trim.check(isMinLength(1), isMaxLength(n))` — verified to trim before
+checking, so `"   "` is still rejected as blank. v4 short-circuits a `.check(…)`
+list on the first failure, so `TimeZone`'s length cap still runs before the ICU
+lookup it exists to protect.
+
+Two test fixes are consequences of phase 3's logger rework rather than of
+Schema. `cire/api`'s unhandled-error log test isolated its line by
+`startsWith('{"')`; the local renderer is indented now, so that filter caught a
+fragment. `Logger.layer` also replaces the whole active set, so there is no
+default logger left emitting the stack dump the filter existed to exclude — the
+whole capture is that one entry, which makes the "no raw SQLite message" check
+strictly stronger. Also drops five `Logger` imports left dead by phase 3.
--- a/.changeset/effect-v4-phase-5-versioned.md
+++ b/.changeset/effect-v4-phase-5-versioned.md
@@ -1,0 +1,15 @@
+---
+"@osn/api": patch
+"@pulse/api": patch
+"@zap/api": patch
+"@shared/observability": patch
+---
+
+Drop five `Logger` imports left dead by the v4 logger rework, and finish the
+Effect v4 migration: with `@cire/api` moved off v3 in the same change, the
+whole monorepo type-checks and passes its tests under Effect v4.
+
+The observability change is the test-only one: `Logger.layer` replaces the
+whole active logger set, so the default logger that used to emit a separate
+"Fiber terminated…" stack dump is gone, and a capture is now exactly the
+entry under test.
--- a/cire/api/src/routes/account-link.ts
+++ b/cire/api/src/routes/account-link.ts
@@ -162,7 +162,7 @@ export const createAccountLinkPostRoute = (
 
         return runCire(
           Effect.gen(function* () {
-            const body = yield* Schema.decodeUnknown(LinkAccountBody)(raw);
+            const body = yield* Schema.decodeUnknownEffect(LinkAccountBody)(raw);
 
             const resolution = yield* Effect.tryPromise({
               try: () => resolveAccount(profileId),
@@ -209,7 +209,7 @@ export const createAccountLinkPostRoute = (
           }).pipe(
             Effect.provideService(DbService, db),
             Effect.catchTags({
-              ParseError: () =>
+              SchemaError: () =>
                 Effect.sync(() => {
                   metricAccountLinkRequest("error");
                   set.status = 400;
--- a/cire/api/src/routes/budget.ts
+++ b/cire/api/src/routes/budget.ts
@@ -120,12 +120,12 @@ export const createBudgetWriteRoutes = (db: Db, osnAuthOptions: OsnAuthOptions) 
             const raw: unknown = await request.json().catch(() => null);
             return runCire(
               Effect.gen(function* () {
-                const body = yield* Schema.decodeUnknown(CreateBudgetItemBody)(raw);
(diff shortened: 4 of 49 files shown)


chav-aniket:
## 📚 Effect v4 stack — 5 of 6 ← **you are here**

| # | PR | Base | CI |
|---|---|---|---|
| **5** | **#912 — Schema: cire/api (last v3 package)** | #911 | 🟢 **first green point** |

**Merge bottom-up into `effect-v4`**, then `effect-v4` → `main`. 1–4 are red by design and must not go through auto-merge (#896); this is where the tree goes green, exactly as the epic predicted.

```bash
```

---

chav-aniket:

Closing rather than merging: GitHub does not auto-close a PR whose commits reach the base by direct push — that only happens for the default branch. This one was green (9/9), but merging it alone would have left the four red PRs beneath it stranded, so the whole stack went in as one fast-forward.

**#902 is not closed by this**; the `effect-v4` → `main` PR carries `Closes #896`…`#903`.

---

Issue #640, opened 2026-08-15
Pulse API: decide what "REST endpoints for third-party consumers" should mean


chav-aniket:
## What and why

`README.md:185` says the backends use "Eden treaty for type-safe internal calls; plain REST for third-party consumers". This issue came from the old TODO list with nothing but that title, and no third-party consumer of Pulse has been named.

Much of it already exists:

- Every `pulse/api` route is plain HTTP and JSON (Elysia): `/events`, `/series`, `/venues`, `/close-friends`, `/me`, `/account` in `pulse/api/src/routes/`. Eden is only a typed client over those same routes.
- The routes have an OpenAPI document, committed at `shared/openapi/pulse.json` and generated by `pulse/api/scripts/generate-openapi.ts`. `pulse/api/src/app.ts:69-83` explains that the live `/openapi` page is served on `local` and `dev` only, because on a public host it maps every route for an attacker.
- Callers authenticate with an OSN access token (a JWT, JSON Web Token) or the Pulse web session cookie; there is no API key or third-party client model.

What is missing is a decision on whether outside apps should call Pulse at all, and if so on what terms.

## Options

**A. Close.** The routes are already plain REST with a committed OpenAPI document. Reopen when a real consumer exists.
Trade-off: costs nothing. Anyone who integrates later depends on routes that were never promised to stay stable.

**B. Define a public, versioned read API.** For example `GET /v1/events` and `GET /v1/events/:id` for public events only, with a stability promise, per-client rate limits, and authentication through the existing OSN OpenID Connect provider (a registered client per consumer).
Trade-off: a public contract to maintain, a new auth path, and rate-limit and abuse work, all before any consumer has asked for it.

**C. Publish the existing OpenAPI document as the contract.** Keep today's routes, mark a subset stable, and ship `shared/openapi/pulse.json` as its documentation.
Trade-off: cheap, but it freezes the internal route shape the web app uses, which then cannot change freely.

## Recommendation

**A.** Pulse has no production domain yet (`PULSE_API_ORIGIN` in `pulse/api/wrangler.toml` is still a placeholder) and no consumer is known. Build B when one shows up, shaped by what it needs.

## Done when

The owner picks an option. For A, close this issue. For B or C, rewrite it with the chosen routes, auth model and a concrete acceptance test.

Issue #639, opened 2026-08-15
Decide what real-time transport Pulse needs beyond Zap's WebSockets


chav-aniket:
## What and why

This came from the December 2025 plan for a single backend ("WebSocket setup for real-time" in `packages/api`). That plan is gone: each product now has its own API on Cloudflare Workers (`pulse/api/wrangler.toml` and `zap/api/wrangler.toml` both set `main = "src/index.ts"`). Nothing in the repo opens a WebSocket today — `zap/api/src/metrics.ts:27-28` defines `zap.ws.connections` and `zap.ws.messages.delivered`, but no route uses them.

Live message delivery for Zap already has its own issue, #564 (WebSocket transport for live message delivery in `@zap/api`), and Pulse event chats run through Zap (`pulse/api/src/services/zapBridge.ts`). So the only part #564 does not cover is live updates for Pulse itself — RSVP counts or event status changing while a page is open. `README.md:187` still lists "WebSockets for live messaging and event updates (planned)".

## Question for the owner

Does Pulse need its own real-time channel?

- **A — No, close this.** Messaging real-time is #564; Pulse pages refresh on load or poll. Recommended: no Pulse feature asks for live updates yet, and pushing one update to every open page on Workers needs a Durable Object to hold the sockets, which is a new subsystem to run.
- **B — Yes.** Rewrite this as a Pulse feature naming the events to push (for example RSVP counts and status changes on the event page), with a Durable Object per event holding the sockets, and move it to `product:pulse`.

## Done when

The owner picks A or B. Under A the issue closes as covered by #564; under B it is rewritten with its own **Done when**.

## Files

- `zap/api/src/metrics.ts:27-28`
- `pulse/api/src/services/zapBridge.ts`
- `README.md:187`

PR #913, opened 2026-09-06
docs(effect): fold the v4 migration's findings into the permanent docs, and tear down the scaffolding

### LINKED ISSUE #903: Effect v4 phase 6: full sweep, merge effect-v4 → main, dev-tier smoke, docs and teardown

chav-aniket:
Phase 6 of the Effect v4 migration (#903), stacked on #912 — the top of the stack.

## Issues

- Closes #903 (full sweep, docs, teardown)
- Part of #895; follows #902 (#912)

> [!important] None of the stack's `Closes` keywords will fire, and that is expected
> GitHub creates the linked-PR relationship only for a pull request targeting the repository's **default branch**. Every PR in this stack targets `effect-v4` or another phase branch, so their keywords record intent and nothing else — which is why #896–#903 still show as open with no linked PR even where the keyword was written.
>
> **The `effect-v4` → `main` PR is what has to carry the full set** — `Closes #896`, `#897`, `#898`, `#899`, `#900`, `#901`, `#902`, `#903` — and it is the one that will actually close them and the epic's sub-issue checklist.

## Merging the stack

`effect-v4` is **fast-forwardable to this branch** — the six phases are strictly linear, 20 commits, no merge commits needed:

```bash
git fetch origin
git push origin origin/claude/effect-v4-phase-6:effect-v4
```

That advances the integration branch in one step and GitHub then marks all six PRs merged, without pressing a merge button on a PR whose CI cannot go green (#906, #908, #909, #911 are red by design per #896). Then open `effect-v4` → `main` with the `Closes` set above.

### Docs — the original scope

`CLAUDE.md` gains an **"Effect v4"** section: the excerpt that makes the two `Effect-TS/skills` unnecessary. It leads with what did *not* change — including `Data.TaggedError`, this repo's entire error vocabulary, which needed zero edits — then the renames, `Cause`/`Runtime`, the Schema surface, the logger contract and layer memoization.

Three wiki pages take the durable findings: `architecture/backend-patterns`, `architecture/schema-layers` (the v4 form table plus the four properties you must know before trusting it), `conventions/testing-patterns`.

**Teardown:** `wiki/runbooks/effect-v4-migration.md`, both Effect skills, `skills-lock.json` and `.agents/` are deleted, along with their `CODEOWNERS` and changeset-allowlist entries. Effect is recorded as **settled for the backend**, not a trial; the frontend stays an open question to revisit at Solid v2.

### Fixes that came out of researching v4 properly

**`fix(osn-api)` — defect containment moved from the allowlist to the runner.** The containment argument written in #908 was **false**. `Cause.squash` returns the *defect* when a cause carries no `Fail`, and a `Data.TaggedError` is an `Error` with a `_tag`, so `Effect.die`, `Effect.orDie` (already used at three sites here) and a bare `throw` in `Effect.sync` all put an allow-listed tag in front of `makeSafeError`. Reproduced, then fixed at `makeAppRunner` via `runPromiseExit` + `Cause.findErrorOption` — one file, all 136 call sites unchanged. Proven non-vacuous: against the old runner the new tests fail *with the secret in the assertion output*.

**`fix(observability)` — the pretty logger is redacted, and a deployed Worker is off it.** #909 claimed `Logger.consolePretty()` was "opaque, no seam to redact through", and traded `local`'s ANSI colour for redaction on that basis. False: v4 exposes the entry on the input side. Colour is back *and* redacted, `redact` no longer throws on cyclic input (it could kill the fiber that logged), and `zap/api/src/index.ts` — a deployed Worker — is off `PrettyLoggerLive`.

**`feat(observability)` — OTLP trace export on both production Workers.** The largest single win of the migration and only possible on v4: `effect/unstable/observability`'s `OtlpTracer` over `FetchHttpClient` is dependency-free and runs on workerd, where `@effect/opentelemetry`'s `NodeSdk`/`WebSdk` do not. **177 `Effect.withSpan` call sites were producing no data.** No call site changed. Bundles verified free of `@opentelemetry/sdk-*`; inert when unconfigured.

**`test(db-utils)` — the search string-math is pinned by property tests**, all three invariants true and each mutation-checked.

### Corrections to my own earlier work

- The **defect containment** guarantee (#908) — false, fixed above.
- **"`consolePretty` is opaque"** (#909) — false, fixed above.
- **Layer memoization** — I wrote here that v4's shared `MemoMap` weakened the rebuild-cost argument for a shared `ManagedRuntime`. It does not: the MemoMap is shared *within one run*, not across separate `runPromise` roots, and a per-request `Effect.provide` is a new root. Measured — 5 provides of a scoped layer give **5 acquires and 5 releases**; a `ManagedRuntime` gives 1 and 0. Corrected in `CLAUDE.md` and `backend-patterns.md`.

## Filed to `xchromo/osn-tracker`

- **#638** — `zap/api/wrangler.toml` sets no tier var for production, so the Worker reads as `local`: the CORS fail-closed guard never fires. **Needs a dashboard check.**
- **#639** — `Logger.tracerLogger` ships raw annotations, messages and causes to Grafana on every tier, bypassing the deny-list.
- **#640** — an untagged defect is answered `400 invalid_request`, hiding server crashes from 5xx alerting.

A fourth candidate was **dropped**: the claim that `cause` leaks a tagged error's fields on the JSON path did not reproduce.

## Verification

`bun run check` 37/37 · `bun run test` 38/38 · `lint` and `fmt:check` clean · `shared/observability` 113 · `osn/api` 1153 · `cire/api` 1752 · `osn/client` 228 · `db-utils` 66 · `redis` 82.

## Deferred

`oxlintrc.json:245` still describes its `.repos` exclusion as *"cloned by the effect-v3-to-v4 skill"*, a skill this PR deletes. Adding that file here flips the changeset gate from `skip` to `required` and this PR ships in no versioned package — so it is a one-line follow-up to `main` once the stack lands.

---


chav-aniket:
**One teardown leftover, deliberately not fixed in this PR.**

`oxlintrc.json:245` still describes its `.repos` exclusion as *"cloned by the effect-v3-to-v4 skill"* — a skill this PR deletes. My teardown sweep covered `wiki/`, `CLAUDE.md`, `README.md`, `.github/`, `scripts/` and `.claude/`, but not root config files. A repo-wide sweep now finds this as the **only** remaining reference to either removed skill.

I wrote the fix (reword generically, as `.gitignore`'s comment was reworded here) and then reverted it, because adding `oxlintrc.json` to this PR flips `scripts/changeset-required.sh` from `skip` to `required` — and this PR ships inside no versioned package, so satisfying that would mean naming an unrelated package in a false changelog entry. That is the same gate that failed on #907, and the allowlist is deliberately mechanical rather than a judgement call.

The other two homes are worse:

- **Phase 1 (#906)**, where the `.repos` line was actually introduced and which already carries changesets — but amending it means rebasing five PRs on top for a comment.
- **`main`** — the line does not exist there yet, so there is nothing to fix until this stack lands.

**So: a one-line follow-up PR to `main` once the stack merges.** Nothing behavioural; the exclusion itself is still wanted (`.repos/` is ~98 MB, and the next upstream checkout would otherwise flood lint).

Flagging rather than fixing so it is a visible decision, not a silent gap.

---

chav-aniket:
## 📚 Effect v4 stack — 6 of 6 ← **you are here** (top)

| # | PR | Base | CI |
|---|---|---|---|
| **6** | **#913 — docs, teardown, follow-up fixes** | #912 | 🟢 |

**Merge bottom-up into `effect-v4`**, then `effect-v4` → `main` as one green PR. 1–4 must not go through auto-merge (#896).

```bash
```

Not part of this stack: **#910**, which targets `main` directly and is independently mergeable.

---

## This PR grew past docs and teardown

Four verified fixes landed here after the description above was written, each researched against the installed `effect@4.0.0-rc.112` and its migration guides.

**`fix(osn-api)`: defect containment moved from the allowlist to the runner.** The containment argument I wrote in #908 was false. `Cause.squash` returns the *defect* when a cause carries no `Fail`, and a `Data.TaggedError` is an `Error` with a `_tag` — so `Effect.die`, `Effect.orDie` (already used at three sites here) and a bare `throw` in `Effect.sync` all put an allow-listed tag in front of `makeSafeError`. Reproduced, then fixed at `makeAppRunner` via `runPromiseExit` + `Cause.findErrorOption`: one file, all 136 call sites unchanged. Proven non-vacuous — against the old runner the new tests fail *with the secret in the assertion output*.

`OpaqueDefect` exposes its `Cause` through a prototype getter over the non-enumerable `Error.cause`, not an own field: a constructor parameter property is enumerable, so `JSON.stringify` would have serialised the whole `Cause` and re-opened the leak through any structured logger.

**`fix(observability)`: the pretty logger is redacted, and a deployed Worker is off it.** Corrects #909's false "no seam" claim; restores ANSI to `local`; `redact` no longer throws on cyclic input (it was able to kill the fiber that logged); `zap/api/src/index.ts` now uses the real observability layer.

**`test(db-utils)`: the search string-math is pinned by property tests.** All three claimed invariants hold, each mutation-checked.

**`chore`: `.agents/`/`skills-lock.json` retired, and Effect is recorded as settled** rather than a trial.

## Filed to `xchromo/osn-tracker`

- **#638** — `zap/api/wrangler.toml` sets no tier var for production, so the Worker reads as `local`: the CORS fail-closed guard never fires and localhost origins are allowed. Pre-existing; **needs a dashboard check.**
- **#639** — `Logger.tracerLogger` ships raw annotations, messages and causes to Grafana on every tier, bypassing the deny-list.
- **#640** — an untagged defect is answered `400 invalid_request`, so a server crash reads as the client's fault and never reaches 5xx alerting.

A fourth candidate was **dropped**: the claim that `cause` leaks a tagged error's fields on the JSON path did not reproduce — `causePretty` renders name, message and stack only.

## Still deferred

`oxlintrc.json:245` describes its `.repos` exclusion as *"cloned by the effect-v3-to-v4 skill"*, a skill this PR deletes. Adding that file here flips the changeset gate from `skip` to `required`, and this PR ships in no versioned package — so it is a one-line follow-up PR to `main` once the stack lands. Explained in an earlier comment.

---

chav-aniket:
**Landed.** `effect-v4` was fast-forwarded to this branch — `804e8d4` → `5ab1418`, 20 commits, no merge commits. All six phase heads verified contained in `effect-v4`.

Closing rather than merging: GitHub does not auto-close a PR whose commits reach the base by direct push (that only happens for the default branch), so all six are being closed by hand with the record intact.

## What is left, in order

**1. Bring `main` in.** `main` has moved **two** commits ahead since `effect-v4` was cut — #890 and #914 — and #914 (`docs: correct six stale CLAUDE.md facts…`) edits `CLAUDE.md` and `wiki/index.md`, both of which this phase rewrote. **Expect conflicts in exactly those two files**, and nowhere else:

```bash
git checkout effect-v4 && git merge origin/main    # resolve CLAUDE.md + wiki/index.md
```

Resolve by keeping both sets of edits — #914's corrections are to rows this phase did not touch, and this phase's §Effect v4 section is new.

**2. Open `effect-v4` → `main`**, and put the whole closing set in its body, because this is the first PR in the migration whose base is the default branch and therefore the first whose keywords will actually bind:

```
Closes #896
Closes #897
Closes #898
Closes #899
Closes #900
Closes #901
Closes #902
Closes #903
```

That closes all eight sub-issues and completes #895's checklist in one merge.

**3. Then:** dev-tier smoke, the `production` environment approval, and the Grafana `logLevel` → `level` migration (out of repo). OTLP trace export is now one env var away — note sampling defaults to `0.1`, so 90% of Worker spans drop until you raise it.

**Also still open:** #910 (the redaction fix to `main`) is independent of this stack, green, and mergeable on its own.

---

Issue #925, opened 2026-09-07
Rewrite the ~285 flagged references that sit inside long comment blocks, then raise house/no-tracker-ref-in-comment to error


chav-aniket:
**What** — Deal with the `house/no-tracker-ref-in-comment` diagnostics that sit **inside a comment block of 10 or more lines**: roughly 285 of the rule's 1,597. Removing the token here is not the job — deciding what the surrounding block was for is. Then, once these and xchromo/osn#924 are both clear, raise the rule from `"warn"` to `"error"` in `oxlintrc.json`.

**Why** — These are the blocks where a real constraint is buried inside an incident narrative. A mechanical strip would take the tag out and leave prose that no longer explains anything, or worse, delete the one sentence that stops someone reintroducing a bug. They need a person deciding what survives, which is why they are split out of #924's ~1,300 safe edits.

**Done when** — `bunx --bun oxlint -c oxlintrc.json .` reports zero `house/no-tracker-ref-in-comment` diagnostics, and the rule is set to `"error"` in `oxlintrc.json`'s top-level `rules` block. This issue owns the severity flip because it finishes last.

**Files where a blind strip destroys real knowledge** — one PR each, human review, no unsupervised agent. For each, write the surviving constraint first as a standalone paragraph, get it reviewed against the code, *then* delete the narrative:

- `osn/api/src/services/recommendations.ts` — D1's 100-bind cap, stated three times as a reference to a tracker issue
- `pulse/api/tests/services/d1ParamCounts.test.ts` — the same cap, five tracker numbers on line 1, and a real test rationale below them
- `cire/invites/src/components/rsvp-saved.ts` — an accessibility floor sized against VoiceOver; "the way to get it back is NOT to lower it" is the whole point of the comment
- `cire/invites/src/lib/z-index.ts` — three ordering invariants plus a Tailwind literal-class constraint
- `cire/api/src/db/d1-session.ts` — `first-primary`, and a module-scope Effect-primitives incompatibility nobody will reconstruct
- `shared/observability/src/logger/redact.ts` — the deny-list's three-part admission test, consulted at the moment someone edits the array
- `cire/host/src/lib/registry-store.ts` — the sibling-cache list is the only record that seven files depend on the contract
- `osn/api/src/lib/ceremony-store.ts`, `cire/invites/src/lib/consent/vendors.ts`, `cire/invites/src/lib/consent/categories.ts`, `pulse/api/src/lib/limits.ts`

**Other categories in this tranche:**

[...]

PR #943, opened 2026-09-08
Report comment-line delta in prep-pr; close out xchromo/osn#925

chav-aniket:
## Summary

What survives from #925, which is now closed as not planned.

The comment-cleanup effort was measured and the measurement was unflattering: #930 added **66 comment lines net** while describing itself as a cleanup, because its instruction — "rewrite each reference to the constraint it stood for" — turns a single-line parenthetical into a paragraph. Nobody was counting, so it took a human reading the PRs to notice.

Two things come out of that:

- **`/prep-pr` now prints a branch's net comment-line delta**, backed by `scripts/comment-delta.ts`. Deliberately a number to look at, not a gate.
- **The four high-risk comment files already cleared** on this branch ship, since the work is done, verified, and removes a real private-tracker citation from a public file. The remaining ~155 files do not — see #925 for why.

#925's own closing comment carries the full argument. The short version: `src` has 40,049 comment lines, 16,997 of them in blocks of ten lines or more, and #925 explicitly forbade touching those blocks. Its scope was 711 tokens on lines that overwhelmingly stay, so it could not reduce comment volume even executed perfectly. The seven-file trial run bought −24 lines and surfaced three defects *in the cleanup itself*, including a rewrite that came out longer than the original and a factual claim that contradicted the code.

## Workspaces affected

Versioned (patch, one changeset): `@osn/api`, `@pulse/api`, `@shared/observability`. Ignored/version-less (patch, separate changeset per convention): `@cire/invites`, `@cire/host`. `scripts/` and `.claude/` are on the changeset allowlist. No behavior changes anywhere — every source edit is comment text.

## Issues

**Closes**

This PR closes no issue. #925 is closed as `not planned` with its reasoning recorded there; this branch is the part of it worth keeping.

**Opened**

None.

### Report the delta, do not gate on it — `convention`

- **Issue** — A comment cleanup that grows comment volume is invisible until a human opens the diff and counts by eye, which is exactly how #930's +66 went unnoticed through an adversarial review pass that only ever asked "was anything lost?".
- **Why** — A hard `net ≤ 0` CI check is the obvious move and the wrong one. It is trivially gamed by joining wrapped lines, and it punishes legitimate work: #929 rewrapped 334 comment lines to relocate 64 misattached doc blocks and nets ≈0 correctly.
- **Solution** — `/prep-pr` prints the figure and reports it in the PR body's test-plan table. No threshold.
- **Rationale** — The number is diagnostic, not normative. A cleanup at +66 is worth a second look; it is not automatically wrong, and no automation can tell the difference.

### Pass the base in rather than resolving it — `bugfix`

- **Issue** — My own first measurement of this reported `+85` comment lines for #937, which was wrong; the real figure is `−40`. I reported that wrong number to the repo owner as evidence against the work.
- **Why** — The base branch had been force-pushed. `git merge-base` then falls back to an older common ancestor, and the base branch's own commits — here #927's rule, its tests and the wiki convention page — land inside the measured diff.
- **Solution** — The script takes the base as an argument and does not resolve it. `/prep-pr` already resolves `$BASE` in Step 0, and the skill now warns that a surprising figure means checking `git log "$BASE"..HEAD` holds only your commits.
- **Rationale** — The failure is silent and directional: it inflates. A measurement tool that quietly overstates the thing it measures is worse than no tool.

**Out of scope** — The ~155 files #925 leaves behind. They are `warn`-level and safe to fix opportunistically; roughly 500 of the 711 diagnostics are plain parenthetical or leading-label strips.

## Test plan

| Gate | Command | Result |
|---|---|---|
| Comment-delta counter | `bun test scripts/tests/comment-delta.test.ts` | 5 pass, 0 fail |
| Script suite | `bun run test:scripts` | 148 pass, 1 skip, 1 fail — the failure is a pre-existing Bun-version TOML assertion in `cire-dev-db-guard.cli.test.ts`, identical on the unmodified branch, and green in CI |
| Self-check on a known figure | `bun run scripts/comment-delta.ts 14c8f9d` on `fix/mechanical-token-cleanup` | reproduces `-40`, matching the hand count |
| `@osn/api` tests | `bun run --cwd osn/api test:run` | 1153 pass, 0 fail |
| `@cire/invites` tests | `bun run --cwd cire/invites test` | 999 pass, 0 fail |
| `@cire/host` tests | `bun run --cwd cire/host test:run` | 1223 pass, 0 fail |
| `@shared/observability` tests | `bun run --cwd shared/observability test:run` | 113 pass, 0 fail |
| `@pulse/api` (`d1ParamCounts`) | `bun run --cwd pulse/api test:run` | 7 pass, 0 fail |
| Rule diagnostics on the four files | `bunx --bun oxlint -c oxlintrc.json --format=json <file>` | 0 |
| Type check | `bun run check` | 38/38 workspaces pass |
| Format | `bun run fmt:check` | passes |
| Changesets | `bash scripts/validate-changesets.sh` | passes |

The four source files were checked against the code rather than against their own prose. Two internal contradictions were fixed on the way: `recommendations.ts` claimed the FOF query binds `profileId` "once" where it binds a fixed number of times (the same file measures three binds a few hundred lines below), and `rsvp-saved.ts` said "five sequential D1 round-trips" twelve lines above another block saying "six serialised".

`redact.ts`'s three-part deny-list admission test, `registry-store.ts`'s seven-file sibling-cache list, and `rsvp-saved.ts`'s "the way to get it back is NOT to lower it" are preserved verbatim — each is the sentence its file exists to protect.

This PR is stacked on #927 (base: `claude/verbose-comments-analysis-8rm8r2`), which must merge first. It is independent of #929, #930 and #937.

---

https://claude.ai/code/session_01VQDjMy2vkipJQiCc9fxRgS

---
_Generated by [Claude Code](https://claude.ai/code/session_01VQDjMy2vkipJQiCc9fxRgS)_

### DIFF
--- a/.changeset/comment-cleanup-residue-cire.md
+++ b/.changeset/comment-cleanup-residue-cire.md
@@ -1,0 +1,19 @@
+---
+"@cire/invites": patch
+"@cire/host": patch
+---
+
+State four comment constraints directly instead of citing a tracker finding.
+
+`rsvp-saved.ts` keeps every load-bearing sentence about the RSVP dwell — the
+knee/floor arithmetic, the VoiceOver and WCAG 2.2 SC 4.1.3 reasoning, and "the
+way to get it back is NOT to lower it" — while losing the `C-L1` citation to
+the private tracker and three blocks of before/after narration. It also fixes
+an internal contradiction: one block said "five sequential D1 round-trips"
+where another twelve lines down says six serialised.
+
+`registry-store.ts` keeps the seven-file sibling-cache list intact, and the
+consent registry's rule about `<head>` resources is now stated as a rule rather
+than as the history of one vendor.
+
+No behavior changes; every edit is comment text.
--- a/.changeset/comment-cleanup-residue-versioned.md
+++ b/.changeset/comment-cleanup-residue-versioned.md
@@ -1,0 +1,17 @@
+---
+"@osn/api": patch
+"@pulse/api": patch
+"@shared/observability": patch
+---
+
+State four comment constraints directly instead of citing a tracker finding.
+
+`recommendations.ts` carried D1's 100-bound-parameter cap as a tracker citation
+three times over; each now states the constraint, and a claim the old text made
+about the query binding `profileId` "once" is corrected — it binds a fixed
+number of times, which is what the file's own measurement a few hundred lines
+down already said. `d1ParamCounts.test.ts` loses five tracker numbers that its
+own six-site index already covers, and `redact.ts` loses a finding tag from its
+fast-path note. The deny-list's three-part admission test is untouched.
+
+No behavior changes; every edit is comment text.
--- a/.claude/skills/prep-pr/SKILL.md
+++ b/.claude/skills/prep-pr/SKILL.md
@@ -213,6 +213,38 @@ the concern to the user and let them decide whether to split. With no user,
 record the observation under `## Decisions` and continue. The wording is in
 `references/workflow-steps.md`.
 
+### Comment-line delta
+
+Print how many comment lines the branch adds and removes:
+
+```bash
+bun run scripts/comment-delta.ts "$BASE"
+```
+
+Report the net figure in the PR body's test-plan table. **This is a number to
+look at, not a gate** — there is no threshold to pass, and no cleanup is
+required to come out negative. `#929` legitimately rewrapped 334 comment lines
+to move 64 misattached doc blocks; a merge-and-rewrap nets near zero and that
+is correct.
+
+It exists because a comment-cleanup branch that *grows* comment volume is worth
+a second look before a human opens the diff, and nobody had been measuring it.
+The failure it catches is real and was measured: a batch instructed to "rewrite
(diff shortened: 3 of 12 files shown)


chav-aniket:
Checked the Tessl review's `referenced_paths_exist` warning — all four flagged paths are extractor false positives, not broken references:

| Flagged | What it actually is |
|---|---|
| `/package.json` | Generic prose mentions (L149, L174, L176) — "that directory's own `package.json`", not a path |
| `config.json` | Inline shell comment naming `.changeset/config.json` (L164) |
| `validate-changesets.sh` | `scripts/validate-changesets.sh` exists and is invoked at L165; L168 is a bare back-reference to it in the same paragraph |
| `wiki/todo/api.md` | A deliberate anti-example (L283) of the bookmark-shaped issue body the section tells you not to write |

No edit needed. The `skill_md_line_count` warning (501 lines) is pre-existing — this PR adds ~15 — and splitting the file is a separate change, not one to fold into a scripts PR.

---

Issue #1050, opened 2026-09-16
Four brand colour tokens have drifted between cire/landing and cire/invites, which claim to be byte-identical


chav-aniket:
`cire/landing/src/styles/global.css` opens its `@theme` block with:

> Brand tokens — kept byte-identical to cire/invites's theme block so the marketing site, the invites it advertises, and the organiser portal all read as one brand. If you change a token here, change it in `cire/invites/src/styles/global.css` too (and vice versa).

They are not byte-identical. Six of the ten shared `--color-*` tokens had diverged when this was found; one was a contrast defect and is fixed, and **four remain**:

| Token | `cire/landing` | `cire/invites` |
|---|---|---|
| `--color-border` | `oklch(85.51% 0.069 144.94 / 0.1)` | `oklch(94.62% 0.0111 89.72 / 0.12)` |
| `--color-error` | `oklch(67.56% 0.1401 21.48)` | `oklch(72% 0.1401 21.48)` |
| `--color-success` | `oklch(75.96% 0.1421 146.94)` | `oklch(64% 0.1421 146.94)` |
| `--color-surface-raised` | `oklch(27.87% 0.0393 149.62)` | `oklch(27.7% 0.0275 152.78)` |

`--color-success` is the widest: 75.96% lightness against 64%, which is a visibly different green on two sites a visitor moves between in one session.

**Why this needs a decision rather than a fix.** Neither value is wrong on contrast grounds — both clear their floors on their own surfaces, which is why the conformance harness passes on both apps. Choosing one is a design call about which site is the source of truth, and that is the repo owner's.

There is a reason to think it is `cire/invites`: its `@theme` literals are held in lockstep with what `derivePalette(evergreen)` emits (`cire/invites/tests/lib/invite-theme.test.ts`), so they are derived from the scheme rather than typed. `cire/landing`'s are typed, and nothing checks them against anything. But that argument does not settle `--color-surface-raised`, where landing's value may be a deliberate choice for a page that is mostly one long scroll rather than a stack of cards.

**Where it is recorded now.** `cire/landing/tests/styles/design-token-contract.test.ts` asserts parity across the whole shared block and carries these four as a `KNOWN_DIVERGENCES` list, so a *new* drift fails immediately. A second test fails if an entry in that list is stale — i.e. if a token is brought back into line without the exemption being removed — so the list cannot quietly outlive the problem.

[...]

Issue #1049, opened 2026-09-16
Design system phase 6: enforce the contract with @shadcn/lint


chav-aniket:
Part of #652. Blocked by phase 5 (#1048) — the rules go to `error` only once the violations are cleared.

**What** — Adopt [`shadcn-ui/lint`](https://github.com/shadcn-ui/lint) (`@shadcn/lint`) as a third oxlint JS plugin, so the token contract and the size scales are enforced rather than merely documented.

**Why** — `wiki/architecture/frontend-patterns.md` §The Rule already forbids copying a class list into a second component, and nothing enforces it; that is why the same outline button exists at three letter-spacings. `@shadcn/lint` is an agent-first linter for Tailwind design systems that ships these rules, runs on **oxlint ≥ 1.80** and **Tailwind v4** — exactly what this repo pins (`package.json:42,46`; `tailwindcss@4.3.3`) — and `oxlintrc.json:279-282` already carries a `jsPlugins` array with two entries, so adoption is a third.

**Rules, sized 2026-09-16 at `b5f9c8f`** —

| Rule | Adopt as | Note |
|---|---|---|
| `no-raw-colors` | `error` | 28 violations. The rule that makes the contract enforceable |
| `no-arbitrary-values` | `error` | ~1,574 reducible occurrences, cleared by phase 5's codemod. Needs an escape for the 184 computed values; must not count the 51 `data-[…]`/`aria-[…]` variants |
| `require-static-classes` | `error` | With a per-directory override — see below |
| `no-unknown-classes` | `error` | Catches the silent-no-CSS failure `frontend-patterns.md:92` describes |
| `no-restyle` | `error`, `allow: ["layout"]` | 102 sites, cleared by phase 5 |
| `no-inline-styles` | **not as shipped** | See below |

*Measured — `grep -rEoh '\b[a-z-]+-\[[^]]+\]|\[[a-z-]+:[^]]+\]'` for arbitrary values; a `(bg|text|border|ring|from|to|via)-(red|…|stone)-[0-9]{2,3}` alternation for raw colours; brace-matched extraction of `style={…}` for inline styles — all over `cire musubi pulse tools osn shared`, `--include='*.tsx' --include='*.astro'`.*

**Done when** — the five rules above fail the build on violation, **each having been seen to fail** (add a violating fixture, watch the non-zero exit, remove it), and `scripts/lint-warning-ceiling.txt` is re-baselined.

[...]

Issue #652, opened 2026-08-15
Design system: one token contract, two component layers


chav-aniket:
**What** — Define a neutral design-token contract in a new `@shared/design-tokens` package — colour **and size scales** — re-key `@osn/ui`'s primitives onto it, and add a product-specific component layer per stack (`@cire/ui` first). Each app maps its own vocabulary onto the contract once. Every shared package, `@cire/ui` included because it is a library, reads only contract tokens through Tailwind utilities namespaced `osn-*`. Application code keeps its own vocabulary.

**Why** — Two token vocabularies are in use (shadcn names in `@musubi/social`/`@pulse/web`, cire's house ramp in `@cire/*`), and `shared/toast/src/toast.css:4-6` had to be written in plain CSS specifically to route around the split. `cire/host` and `cire/vendor` carry ~90 token names in common, duplicate copies of `Button`/`Card`/`Field`/`Notice`/`EmptyState`, and a duplicated contrast test. `cire/invites` and `cire/landing` have no component layer at all. `wiki/architecture/frontend-patterns.md` §The Rule already forbids this and nothing enforces it.

The sharpest measure is the size drift:

| Utility | Uses | **Distinct values** |
|---|---|---|
| `text-[…]` | 956 | **93** |
| `tracking-[…]` | 275 | 18 |
| `leading-[…]` | 110 | 12 |
| `max-w-[…]` | 92 | 33 |

Ninety-three distinct font sizes, nine of them between `0.7rem` and `0.95rem`. `max-w-[640px]` and `max-w-[40rem]` are the same width written twice. Of 1,758 arbitrary-value occurrences, **1,574 are static and collapse onto a scale**; 184 are computed (`var`/`calc`/`clamp`/gradient) and legitimate.

Component duplication, same date: `ui/` diffs of Button 89 · Card 49 · Field 138 · Notice 24 · EmptyState 14 lines; 9 hand-rolled `fixed inset-0 z-…` overlays; 186 class attributes over 120 chars; the same outline button at three letter-spacings (`ui/Button.tsx` `0.1em`, `PreviewInviteButton.tsx:93` `0.12em`, `invite/InviteBuilder.tsx:995` `0.08em`).

[...]

PR #1060, opened 2026-09-17
Design system: one token contract, three component layers (phases 0–5)

### LINKED ISSUE #1048: Design system phase 5: adoption — convert raw elements, apply the scale mapping, clear the override sites
Part of #652. Blocked by phase 4 (#1047). **Split into one sub-issue per app when it is reached** — the six below are independent and fan out.

**What** — Per app: convert raw `<button>` to `<Button>`, hand-rolled overlays to the shared `Modal`, apply the phase 1 size-scale mapping by codemod, and resolve `no-restyle` violations by adding component variants.

**Why** — The abstractions exist after phases 3 and 4; this is where the duplication actually goes away. Measured 2026-09-16 at `b5f9c8f`: 186 class attributes over 120 chars, 21 files hand-rolling the gold outline button, 9 hand-rolled overlays, ~1,574 reducible arbitrary values, 102 component override sites.

**Apps** — `cire/host` · `cire/invites` · `cire/vendor` · `cire/landing` · `pulse/web` · `musubi/social`

**Done when**, per app —

- Raw `<button>` count in that app's `.tsx` falls to deliberate exceptions only.
- Its overlays use the shared `Modal`.
- Its arbitrary values are on the scales.
- It has no `no-restyle` violations.
- Its browser tier passes — that is what proves a substitution was value-preserving, since jsdom computes no styles.

**Named first targets** —

- `cire/host/src/components/PreviewInviteButton.tsx:93` — needs a new bar-height `size` variant (`h-9`, to sit inside the top bar rather than set its height) plus the icon-collapse behaviour. Everything else is already `BASE`.
- `cire/host/src/components/invite/InviteBuilder.tsx:995` — straight `<Button variant="outline" size="sm">`, no new variant. It is `VARIANT.outline` written longhand, minus `hover:border-gold`.
- `cire/invites` — 26 raw buttons and no component layer; the largest single win.

Those two host buttons plus `cire/host/src/components/ui/Button.tsx` are the same outline button at three letter-spacings (`0.12em`, `0.08em`, `0.1em`), which is what this whole epic is about.

**Adding a variant always beats a call-site override.** `cire/host`'s Button docblock already says so and there is no `tailwind-merge` in that package (`Button.tsx:22-28`).

**The scale migration is a codemod, not hand-editing.** Phase 1 publishes a closed old→new mapping table; write the transform once and run it per app, then read the diff. Two things it must not touch: the 184 **computed** values (`var`/`calc`/`clamp`/gradient — overwhelmingly `cire/invites`' runtime heading scale) and the 51 `data-[…]`/`aria-[…]` **variants**, which are not arbitrary values at all.

**The 102 `no-restyle` sites are a migration into variants.** Samples: `<Input class="bg-muted/50 text-body">`, `<Button variant="ghost" size="sm" class="text-destructive">`, `<Badge variant="outline" class="text-[10px] tracking-wide uppercase">` — that last one is a missing `size` on `Badge`. Layout overrides (`w-full`, `self-start`, grid placement) stay legal under `allow: ["layout"]`.

**Notes** — `wiki/architecture/frontend-patterns.md:54` §The Rule offers "lift a token into `lib/ui.ts`" as the alternative to a component; note that `pulse/web/src/lib/ui.ts` is the **only** such file in the repo, so that half of the rule is pulse-scoped today.

**Changesets:** per app; `@cire/*` are version-less and never share a changeset with a versioned package.

### LINKED ISSUE #1052: Design system phase 5b: cire/host adoption — 98 raw buttons, 8 overlays, 645 arbitrary values
Part of #1048. Needs the codemod from #1051.

The largest of the six by every measure. **Measured 2026-09-16** at `1a1cecc`, over `cire/host/src/**/*.{tsx,astro}`:

| | |
|---|---|
| Raw `<button>` | 98 |
| Files with a hand-rolled `fixed inset-0` overlay | 8 |
| Reducible arbitrary values | 645 (1 computed, leave it) |
| `class` attributes over 120 chars | 67 |
| Shared-component override sites with a non-layout class | 31 |

**Done when** —

- Raw `<button>` count falls to deliberate exceptions, each carrying a comment saying why it is not `@cire/ui`'s `Button`. `cire/invites/src/components/RsvpModal.tsx` is the worked example of an exception that earns its place: it marks a *confirmed* state with `aria-disabled`, and `Button` fades anything carrying that attribute.
- Its overlays use `@osn/ui`'s `Modal`. That component is a native `<dialog>` in the top layer, so each conversion also deletes a `z-index` and a portal — and removes the class of bug that put cire's RSVP toast under the sheet it fired beneath, where a `transform` on any ancestor traps a `position: fixed` child.
- Arbitrary values are on the scales, via #1051's codemod rather than by hand.
- The 31 override sites become component variants or are shown to be layout-only. `<Td class="text-center">` is the commonest and is a missing `align` prop that `Th` already has.
- `bun run --cwd cire/host test:browser` passes. That tier is what proves a substitution was value-preserving, since happy-dom computes no styles and a class-list assertion would pass against an unstyled element.
- `bash scripts/guard-bundle-size.sh cire/host` passes. It sits at 231,558 bytes against a 238,617 ceiling — **~7.1 KB of headroom**, so this work has to come out net-neutral or better. It should: every converted button deletes a long class string.

**Named first targets**, both of which are the same outline button written longhand at a different letter-spacing:

- `src/components/PreviewInviteButton.tsx:93` — needs a bar-height `size` variant (`h-9`, to sit inside the top bar rather than set its height) plus the icon-collapse behaviour. Everything else is already `BASE`.
- `src/components/invite/InviteBuilder.tsx:995` — a straight `<Button variant="outline" size="sm">` with no new variant. It is `VARIANT.outline` written out, minus `hover:border-gold`.

**Adding a variant always beats a call-site override.** There is no `tailwind-merge` in this package, so two competing utilities resolve by their order in the generated stylesheet rather than by the order they appear in `class` — a call-site override of a component default is a coin flip, not a decision. `@cire/ui`'s `Button` docblock says so at the source.

### LINKED ISSUE #1053: Design system phase 5c: cire/invites adoption — 20 raw buttons, 2 overlays, 303 arbitrary values
Part of #1048. Needs the codemod from #1051.

**Measured 2026-09-16** at `1a1cecc`, over `cire/invites/src/**/*.{tsx,astro}`:

| | |
|---|---|
| Raw `<button>` | 20 (was 26; eight became `<Button variant="cta">` in #1047, one is a documented exception) |
| Files with a hand-rolled `fixed inset-0` overlay | 2 |
| Reducible arbitrary values | 303 |
| **Computed values — must not touch** | **24** |
| `class` attributes over 120 chars | 67 |
| Shared-component override sites with a non-layout class | 0 |

**This app carries the codemod's hardest case.** Its 24 computed values are the per-wedding heading scale — `text-[calc(clamp(2rem,5vw,3rem)*var(--invite-heading-scale,1))]` and relatives — where the bracket body depends on a custom property an organiser sets. Snapping one to a scale step deletes a feature people are paying for. Read this app's codemod diff more carefully than any other's, and check the skip report names all 24.

**The remaining 20 buttons are mostly not `Button`-shaped**, which is why #1047 stopped at eight. They are icon-only closes (`AnimatedModal`, `DemoModal`), inline text links styled as buttons (`LoginSection`'s privacy-choices, `ConsentGate`'s), and the RSVP attendance toggles, whose classes are driven by a selected state. Each wants a decision rather than a substitution:

- The toggles are a **segmented control**, not three buttons. They appear in `RsvpModal` and again in `cire/landing`'s demo with the same classes, so they are a component.
- The icon-only closes want `<Button size="icon">`, which exists.
- The text links want to be `<a>` or a `variant="link"`, and today they are `<button>` with no border — which is the right element for an in-page action and the wrong one for navigation.

**Done when** —

- Raw `<button>` falls to deliberate exceptions, each with a comment saying why.
- Its two overlays use `@osn/ui`'s `Modal`, deleting a `z-index` and a portal each. `lib/z-index.test.ts` guards the numbers today because a jsdom test cannot check that a popover actually paints above the modal; a top-layer dialog makes that class of bug impossible rather than guarded.
- Arbitrary values are on the scales, with the 24 computed ones untouched and visible in the codemod's skip report.
- `bun run --cwd cire/invites test:browser` passes, including `tests/styles/token-contract.browser.test.tsx`, which is what proves the contract still reaches the pixel through an organiser's runtime palette.
- `bash scripts/guard-bundle-size.sh cire/invites` passes — 166,341 bytes against a 175,000 ceiling.

### LINKED ISSUE #1054: Design system phase 5d: musubi/social and pulse/web adoption — 54 override sites, 100 arbitrary values
Part of #1048. Needs the codemod from #1051.

The two Solid apps are one sub-issue rather than two: between them they have 100 arbitrary values and no overlays, and what they actually carry is the **override problem** — 54 of the repo's component-override sites, against 31 in `cire/host` and none in the other three.

**Measured 2026-09-16** at `1a1cecc`:

| | `musubi/social` | `pulse/web` |
|---|---|---|
| Raw `<button>` | 6 | 24 |
| Hand-rolled `fixed inset-0` overlays | 0 | 0 |
| Reducible arbitrary values | 3 | 97 (2 computed) |
| `class` attributes over 120 chars | 2 | 16 |
| Shared-component override sites with a non-layout class | 34 | 20 |

**Every override is a missing variant.** Three patterns cover most of them, and each names the variant to add rather than the class to delete:

- `<Button class="text-body rounded-pill">` — musubi's commonest, ×34 between its variants. musubi's buttons are pill-shaped and its type scale is four steps, so this is a `size` the shared `Button` does not have rather than a per-call-site correction.
- `<Badge class="tracking-wide uppercase">` — pulse's commonest. A missing `size` on `Badge`, exactly as #1048 called it.
- `<Input class="bg-muted/50">` — a filled input, which is a variant.

**Done when**, per app —

- Override sites carry only layout classes (`w-full`, `self-start`, grid placement), which stay legal under `no-restyle`'s `allow: ["layout"]`. Everything else is a variant on the shared component.
- Raw `<button>` falls to deliberate exceptions, each with a comment.
- Arbitrary values are on the scales, via #1051's codemod.
- `bun run --cwd musubi/social test:browser` and `bun run --cwd pulse/web test:browser` pass. Those tiers exist as of #1045 and already assert that the contract reaches the pixel in each app; they are what proves a variant substitution was value-preserving, since happy-dom computes no styles.

**Adding a variant to `@osn/ui` changes both apps at once**, which is the point and also the risk: a new `size` on `Button` has to leave every existing call site rendering identically. The browser tier in both apps is the check, and `bun run --cwd osn/ui test:browser` is where a new variant earns its own coverage.

**Note** — `wiki/architecture/frontend-patterns.md` §The Rule offers "lift a token into `lib/ui.ts`" as the alternative to a component. `pulse/web/src/lib/ui.ts` is the only such file in the repo, so that half of the rule is pulse-scoped today; it is not a pattern to spread to musubi without deciding to.

### LINKED ISSUE #1055: Design system phase 5e: cire/vendor and cire/landing adoption — 10 raw buttons, 139 arbitrary values
Part of #1048. Needs the codemod from #1051.

The two small cire surfaces, together because neither is more than an afternoon.

**Measured 2026-09-16** at `1a1cecc`:

| | `cire/vendor` | `cire/landing` |
|---|---|---|
| Raw `<button>` | 6 | 4 (was 7; three became `<Button variant="cta">` in #1047) |
| Hand-rolled `fixed inset-0` overlays | 0 | 1 |
| Reducible arbitrary values | 46 (2 computed) | 93 (13 computed) |
| `class` attributes over 120 chars | 1 | 7 |
| Shared-component override sites with a non-layout class | 0 | 0 |

**`cire/vendor` is the counterexample the epic was built on** — same Tailwind, same Astro, 6 raw buttons and one long class attribute, against `cire/host`'s 98 and 67. It is the smallest of the six and it is the one to do first, because it is where a mistake in the codemod is cheapest to spot.

**`cire/landing` has the tightest bundle in the repo.** 185,488 bytes against a 189,324 ceiling — **~3.8 KB of headroom**, the least of any app. Its 13 computed values are the vine backdrop's typed custom properties (`--p`, `--a0`, `--a1`), which the codemod must leave alone: they are animated per scroll position and a snapped value is a frozen animation.

Its remaining overlay is `demo/DemoModal.tsx`, which hand-rolls a `fixed inset-0` backdrop and an icon-only close. Both have shared answers now — `@osn/ui`'s `Modal` (a native `<dialog>`, so the conversion deletes a `z-index` and a portal) and `<Button size="icon">`.

**Done when**, per app —

- Raw `<button>` falls to deliberate exceptions, each with a comment saying why.
- `cire/landing`'s overlay uses `@osn/ui`'s `Modal`.
- Arbitrary values are on the scales, with the computed ones untouched and visible in the codemod's skip report.
- `bash scripts/guard-bundle-size.sh cire/vendor` and `… cire/landing` pass. Vendor sits at 77,222 against 81,644 and landing at 185,488 against 189,324, so both need this work to be net-neutral or better. It should be: every converted button deletes a long class string.
- `cire/vendor` has no browser tier, so its check is the unit suite plus the bundle guard. If a conversion there turns out to need a painted-colour assertion, that is a reason to give it one rather than to assert the class list and hope.

chav-aniket:
## Summary

Two token vocabularies were in use across ten frontends — shadcn names in `@musubi/social` and `@pulse/web`, cire's house ramp in `@cire/*` — and `shared/toast/src/toast.css` had been written in plain CSS specifically to route around the split. This branch replaces both with one neutral contract in `@shared/design-tokens`: colour roles plus type, tracking, leading and measure scales, namespaced `ui-*` so it is a library-facing API. Every shared package reads only contract tokens; each app maps its own vocabulary onto them once. StyleX was evaluated and rejected — six of ten frontends are Astro, where it cannot be used at all.

The component half is the same argument. The old `@osn/ui` is re-keyed, gains ten primitives lifted out of cire, and then **splits in two**: `@shared/ui` holds the primitives and `@osn/auth-ui` the auth views. A new version-less `@cire/ui` holds the house style that is genuinely cire's, and the apps adopt all three. The brief's own example is fixed: `PreviewInviteButton` and `InviteBuilder`'s Preview control are now the same `<Button variant="cta">` rather than two hand-written class lists at two letter-spacings.

- **Sizes became part of the contract, not just colour.** 93 distinct font sizes existed; a codemod (`scripts/codemod-scale.ts`) moved 1,021 arbitrary values onto the scales across all six apps and printed the 37 computed ones it skipped, by name.
- **The contract surfaced 13 pre-existing contrast defects**, the worst being cire's primary button at 2.42:1. Eleven are fixed here; two are waived and tracked.
- **Nine hand-rolled `fixed inset-0` overlays became one `Modal`** on the platform's `<dialog>`. That moved the guest site's sheets into the **top layer**, which paints above every stacking context and makes the rest of the document `inert` — and that had consequences outside the sheets, covered under Decisions.
- **`@osn/ui` no longer exists.** OSN is the spec and Musubi is our implementation of it, and a `Button` is not something an independent implementation has to spell our way. The primitives are `@shared/ui`; the auth views, each the client half of a named ceremony, stay in `osn/` as `@osn/auth-ui`. The token prefix moved with them — `--osn-surface` is now `--ui-surface`.
- **The design system is now explorable.** Fourteen stories in `@tools/lab` under `design-system/`, measuring what the browser painted rather than what the stylesheet claims — and they immediately found a bug that had been re-theming the whole lab.
- `@shadcn/lint` enforcement (#1049, the last of eight phases) is **not** in this PR. It is #1062, stacked on this one so it can merge on its own schedule.

An adversarial review of the whole branch ran after this PR opened; its findings and their dispositions are [in a comment below](https://github.com/englishstventures/osn/pull/1060#issuecomment-5707167090), and the two it turned into code changes are Decisions here.

## Workspaces affected

`@shared/design-tokens` (new), `@shared/color` (new), `@shared/ui` (new — the primitives out of `@osn/ui`), `@osn/auth-ui` (renamed from `@osn/ui`), `@shared/toast`, `@cire/ui` (new), `@cire/host`, `@cire/vendor`, `@cire/invites`, `@cire/landing`, `@cire/theme`, `@cire/api`, `@musubi/social`, `@pulse/web`, `@osn/api`, `@tools/lab`, `@tools/metrics`, plus `scripts/` and `wiki/`.

42 changesets. `@shared/*`, `@musubi/social`, `@pulse/web` and `@tools/lab` take `minor`, `@osn/auth-ui` `major` (every subpath it exports moved); the rest `patch`. The version-less `@cire/*` packages never share a changeset file with a versioned one — `scripts/validate-changesets.sh` enforces that and caught one mixed file during the branch (`813345b`).

## Issues

**Closes**

| Issue | What |
|---|---|
| #1048 | Design system phase 5: adoption — convert raw elements, apply the scale mapping, clear the override sites |
| #1052 | Design system phase 5b: cire/host adoption |
| #1053 | Design system phase 5c: cire/invites adoption |
| #1054 | Design system phase 5d: musubi/social and pulse/web adoption |
| #1055 | Design system phase 5e: cire/vendor and cire/landing adoption |

Closes #1048
Closes #1052
Closes #1053
Closes #1054
Closes #1055

**Opened**

| Issue | What |
|---|---|
| #1061 | An undecided guest has no consent affordance while a details sheet is open — `needs:decision`, three options |
| #1050 | Four brand colour tokens have drifted between cire/landing and cire/invites, which claim to be byte-identical |
| xchromo/osn-tracker#653 | — |

#652 stays open: its eighth sub-issue, #1049, is #1062.

### StyleX rejected, CSS Modules rejected, Tailwind v4 kept — `approach`

- **Issue** — the brief asked whether the repo should move from Tailwind to StyleX, with CSS Modules considered too.
- **Why** — the answer decides every later phase, and a wrong one is a rewrite of ten frontends.
- **Solution** — both rejected, and the work redirected at the real cause.
- **Rationale** — six of ten frontends are Astro and StyleX cannot be used in `.astro` files; there is no official Vite plugin and the repo is on rolldown-vite; StyleX bans the descendant and sibling selectors in use (49 `data-[…]`, 20 `group-`, 5 `[&…]`) and the 81 named container-query variants that drive the layout. **CSS Modules are closed on the same ground and one of their own**: they do not compose with Tailwind seamlessly, so adopting them means running two styling systems side by side, and in `.astro` files they would still cost the `data-[…]`, `group-` and container-query variants the layout is built on. Where a shared package genuinely needs a plain stylesheet the repo already has the answer and it is not CSS Modules — `@shared/toast` ships one, themed through custom properties the app maps. More to the point, **Tailwind was never the cause of the duplication**. The abstraction already existed and was not adopted — `cire/vendor` is the counterexample, using the same primitives with almost no hand-written classes.

### The contract is library-facing, and apps keep their own vocabulary — `approach`

- **Issue** — a shared component cannot know whether its host calls the primary ground `--primary` or `--gold`.
- **Why** — every attempt to share a component across the two vocabularies had previously ended in `!important` or a plain-CSS workaround.
- **Solution** — `ui-*` tokens under `@theme inline`, written by shared packages only; each app declares the mapping once.
- **Rationale** — `inline` means `bg-ui-surface` emits `var(--ui-surface, <fallback>)` resolved **at the element**, so a subtree that redefines a token is followed correctly and no theme scope has to be pinned to `<html>`. App code is untouched: every `bg-card` and `bg-gold` stays as written. Written up in the new `wiki/architecture/design-tokens.md`, and now demonstrated live — see the lab Decision below.

### `@osn/ui` splits into `@shared/ui` and `@osn/auth-ui` — `approach`

- **Issue** — one package held both the `Button`/`Card`/`Modal` primitives that five products render and the `SignIn`/`PasskeysView`/`StepUpDialog` views that only the identity app runs, under a name asserting all of it was OSN's.
- **Why** — `wiki/architecture/osn-and-musubi.md` states the discriminator: *if an independent implementation must use the same string to interoperate, it is OSN; otherwise it is not.* A `Button` fails that test outright. The page knew it, and carved the package out by hand — "it keeps its name because `@pulse/web` and `tools/lab` consume it as well" — which is a reason not to call it Musubi's, not a reason to call it OSN's.
- **Solution** — `@shared/ui` for the primitives and `cn()`/`clsx()`; `@osn/auth-ui` for the auth views, which stay in `osn/` and now depend on `@shared/ui` like any other consumer. Subpaths flatten: `@osn/ui/auth/SignIn` → `@osn/auth-ui/SignIn`, `@osn/ui/ui/button` → `@shared/ui/ui/button`. The hand-written carve-out in `osn-and-musubi.md` is replaced by the rule applied.
- **Rationale** — the auth views are the interesting half, because they *look* like product code. What keeps them in `osn/` is that each is the client half of a named ceremony in the spec, so an independent implementation rebuilding those endpoints would rebuild these screens too. The token prefix moved in the same pass for the same reason: a prefix naming the system a component is *not* part of was the last thing asserting the old shape. The rename was driven by an **explicit allowlist of the contract's own token suffixes**, not a blanket `osn-` → `ui-` substitution — the same compound shape carries `osn-access`, `osn-step-up`, `osn-kid`, `osn-pairwise-salt` and the `osn-api`/`osn-social` Cloudflare project names, so a missed token fails safe by staying `osn-` where a mangled audience string would not. That allowlist did miss one shape, the glob form `--osn-*` used in prose; twelve sites including `tokens.css`'s own integration example, swept later in the branch. `ProfileOnboarding`'s `localStorage` key is deliberately left spelled the old way: every browser that has dismissed that prompt holds it under the old string.

### The design system is explorable, and a story had been re-theming the lab — `approach`

- **Issue** — the contract could be read but not seen. Fourteen new stories in `@tools/lab` under `design-system/` fix that: colour roles shown in what they become rather than as chips, the type/tracking/leading/measure scales stacked so a scale reads as one, radius, elevation, the focus ring, motion that actually moves, and three token mappings side by side.
- **Why** — the first screenshot of the new type-scale story in dark mode came back dark-grey on near-black. `cire/host/src/components/ModuleSidebar.story.tsx` imports `../styles/global.css` for the portal's shapes; that file declares a global `:root` mapping `--ui-*` onto cire's ramp, and the lab builds its sidebar by importing **every** story module — so the block landed on the page whatever story was open, and loading after the lab's own stylesheet, it won. Every `@shared/ui` component in every lab story had been rendering in cire's light palette, with `--ui-ink` deaf to the light·dark toggle, because cire's dark ramp keys off `data-theme` rather than the `.dark` class the lab sets.
- **Solution** — the import is removed and the rule stated in the file: a story must not import a stylesheet that declares a global `:root`. The stories themselves read computed style through a `MutationObserver` on `<html>`'s `class`, so every number beside a swatch is the current theme's real paint; group membership derives from `CONTRACT_COLOR_TOKENS` and `CONTRACT_SCALES` rather than a hand-typed list, so a token added to the contract lands in a group or in a visible `unsorted` row.
- **Rationale** — **every gate in this repo was green with that bug in place**, and had been since the story was added. `check`, the full suite, `lint`, `fmt:check` — none of them looks at a pixel outside the browser tier, and the lab has no browser tier. The story's own comment even described the effect, understated: it claimed to "re-theme the chrome around the story" when the reach was the whole lab, permanently. It took a screenshot to see. That is also the argument for the contrast story existing: it prints a live ratio for all 43 pairs `contrastPairs()` generates and marks musubi's two waived pairs as failing, because a waiver suppresses an assertion rather than the defect.

### Thirteen contrast defects the contract exposed — `approach`

- **Issue** — deriving each app's mapping meant computing the ratios, and eleven pairs failed WCAG 2.2.
- **Why** — cire's primary button measured **2.42:1** with a light label and **1.80:1** on dark hover. `--ring` in pulse measured 2.38:1, below the 3:1 floor SC 1.4.11 sets for a focus indicator — a ring nobody can see removes the feature outright for keyboard users.
- **Solution** — fixed in the mapping, with the measured ratio in a comment beside each changed value, and a conformance harness in `@shared/design-tokens` that every app now calls from its own suite.
- **Rationale** — these were live defects found by arithmetic, not by the refactor. Two remain open because the fix is a design choice rather than a derivation (tracked in `osn-tracker#653`), and the lab's contrast story now shows both of them failing rather than letting the waiver hide them.

### `Modal` on `<dialog>`, with a JavaScript-deferred exit — `approach`

- **Issue** — nine hand-rolled `fixed inset-0` overlays, each one animated ancestor away from the `transform`-traps-`position: fixed` bug that has already cost this repository a mispositioned toast.
- **Why** — the top layer makes that whole class of bug impossible rather than avoided by convention, and `showModal()` supplies the focus trap, Escape, background inertness and the backdrop that five of the nine never had at all.
- **Solution** — one `Modal`, entry in pure CSS via `@starting-style`, exit deferred from JavaScript until `getAnimations({ subtree: true })` settles.
- **Rationale** — the CSS-only exit looks like it should work and does not: the platform's answer is the `overlay` property with `transition-behavior: allow-discrete`, and `overlay` is **Chrome and Edge only**. On cire's guest site most traffic is mobile Safari, so a CSS-only exit means the dialog blinking away for nearly everyone who sees it. Kobalte was rejected for these two apps on a measurement, not a preference: `@kobalte/core/dialog` costs 15.3 KB gzip against ~11.7 KB of budget headroom. That argument now lives in `wiki/decisions/native-dialog-over-kobalte.md` rather than in 38 lines of comment above the component.

### Nine `Modal` call sites were silently overruled by the component — `approach`

- **Issue** — `Modal`'s defaults are `base:`-prefixed (`:where(…)`, zero specificity) so that a caller's **plain** utility wins. A `base:` one *ties*, and a tie resolves by the order Tailwind emitted the rules in — not the class-attribute order, and nothing visible at the call site.
- **Why** — the consent preferences dialog was the worst of the nine: measured at 480px instead of the 512px it asked for, and painted on `--ui-surface-raised` instead of the page ground, which put the panel on the same colour as the category rows inside it. Every string assertion in the suite stayed green.
- **Solution** — all nine converted to plain utilities, a browser test on the consent dialog's surface, width and row contrast, and the rule written up in `component-library.md`. The durable guard is `house/no-base-variant-at-call-site` in #1062, which found forty more sites.
- **Rationale** — found by the review, not by CI, and invisible to anything that reads a class attribute. It is the strongest argument in the branch for the browser tier existing at all.

### The top layer broke two things outside the sheets — `approach`

- **Issue** — a `showModal()` dialog paints above every stacking context **and** makes every node outside it `inert`. `AddToCalendar`, portalled to `<body>` at `z-110`, became painted-but-dead; the RSVP save toast, which fires while the sheet is still open for its dwell, could no longer out-number the sheet at any `z-index`.
- **Why** — the calendar menu is opened from inside the sheet, so "visible and unclickable" is the exact #203 bug report arriving through a different door. The save toast is the only confirmation a partial save gets.
- **Solution** — `AddToCalendar` renders **in place** and shows itself as a `popover`, which does the job the portal was there for while staying a descendant of the dialog and therefore reachable. `@shared/toast` gains `topLayer`, which shows the container as a popover as the first toast arrives, so it enters the top layer after any open dialog.
- **Rationale** — both were found by the browser tier, not by review, and neither is visible to a jsdom test or to a `z-index` guard. `Z_LAYER.MODAL` is deleted rather than left stale: an entry claiming to rank against a top-layer dialog is a promise the scale cannot keep. The **consent banner** is the third thing the top layer changed — it is not a dialog, so while a sheet is open it is painted underneath and inert. That is documented rather than fixed, and the gap it leaves is #1061.

### The toast is painted above a sheet but is `inert` under one — `approach`

- **Issue** — `topLayer` fixes the paint order. It cannot fix reach: only a descendant of the open dialog escapes `inert`, and a toaster mounted once at the page root can never be one.
- **Why** — while a sheet is open the toast is seen and nothing more. Its close button does not respond and assistive technology does not announce it.
- **Solution** — accepted, and documented on `ToasterProps.topLayer` and in `wiki/systems/toast.md` in the form "it buys paint, not reach", with the rule that a toast raised over a modal must be a confirmation the dialog itself also states. cire's RSVP sheet flips its own button to "Saved" for exactly this reason.
- **Rationale** — **This is the weakest point in the branch and a reviewer should push on it.** The review agreed the limitation is acceptable as documented but proposed a better fix I have not taken: raise the toast when the dwell timer closes the sheet rather than when the save lands, which would make it both hit-testable and announced. That is a product timing change to the RSVP confirmation, not a refactor, and it needs a dwell figure someone has looked at. The paint-order half is reasoned from the spec and confirmed indirectly (a popover shown after the dialog hit-tests above it, `cire/invites/tests/lib/z-index.browser.test.tsx`); the `inert` half is confirmed directly.

### Tests that moved tiers rather than being deleted — `approach`

- **Issue** — jsdom implements no part of `<dialog>`: `showModal` is `undefined`, nothing is focused on open, and every box measures zero. Escape, the backdrop, modality and the focus restore became unassertable in the fast tier.
- **Why** — a test that silently becomes vacuous is worse than one that is deleted.
- **Solution** — the fast tier asserts what the app owns by dispatching the `close` event the platform would have fired, and that the scrollport carries `autofocus`; the gestures themselves moved to new browser files. `Modal` also degrades to a non-modal dialog under jsdom rather than throwing, and keeps its reduced-motion coverage.
- **Rationale** — four latent test bugs fell out of the move and are fixed here: two browser files were measuring the previous test's still-open dialog because nothing called `cleanup()`; one was reading `getBoundingClientRect()` during the entry transform, which reports the box scaled by 0.98; and two Escape tests used `requestClose()`, which goes through the close-watcher budget and quietly stops closing anything once a suite has opened a few dialogs without user activation. Both now press Escape for real through `userEvent.keyboard`.

### The browser tier runs serially — `approach`

- **Issue** — the browser job failed on three of this branch's six CI runs, on a different package each time (`@musubi/social`, the old `@osn/ui`), always after `Port 63315 is in use, trying another one...` and a navigation that never completed, with every test that ran passing.
- **Why** — this branch took `test:browser` from two projects to six, so six Vite servers and six Chromiums start at once on a four-core runner. A failure that moves between packages is contention, not a test.
- **Solution** — `--concurrency=1` on the root `test:browser` script, matching what `test:d1` already carries, with the reason in `turbo.json` beside the task.
- **Rationale** — it costs wall-clock and buys a browser tier that means something. An intermittent red that lands somewhere different each time teaches everyone to re-run rather than to look, and the browser tier is the only place several of this branch's guarantees are checkable at all.

### Nine decision records moved out of source files into `wiki/decisions/` — `approach`

- **Issue** — `modal.tsx` carried 38 lines arguing for the platform `<dialog>` over the Kobalte one sitting beside it in the same package. It was not the only one: 571 lines across seven files were doing the same job.
- **Why** — a reader consults that argument once and then never again, but every future reader of the file pays for it, and so does every agent that loads it.
- **Solution** — a new `wiki/decisions/` directory with eight pages plus a README that draws the line, and each source file left with the sentences that state what the code guarantees **now** plus a `@see` to the page by repo path.
- **Rationale** — by repo path and not `[[wikilink]]`, because a wikilink resolves in neither GitHub nor an editor. The line the README draws: a choice between named alternatives goes in `wiki/decisions/`; a comment stating a current guarantee stays in the code; a deferral is an issue. The D1 page is deliberately short because `wiki/systems/d1-read-replication.md` already carries that argument — a second full page would be a duplicate that drifts. Every measured figure moved with its existing marker rather than being re-measured, and no executable line changed: the seven source diffs are comments only.

**Out of scope** — `osn-tracker#653`. #1050. #1061. #1049 is #1062. Two adoption items are deferred on design grounds and named in `NEW-FEAT.md`: the `Badge` eyebrow treatment (5 sites, needs a second CVA axis) and `AvatarFallback` sizing (19 sites, a non-constant ratio). `musubi`'s `ThemeToggle` and `ProfileSwitcherDialog` rows look alike and are not — folding them together would mean a prop per difference. `cire/invites/src/components/InviteClosing.tsx` carries two review-finding tags in comments (`house/no-tracker-ref-in-comment` warns on them); restating them as durable reasons is a deliberate edit rather than a side effect of moving prose, and is left alone here.

## Test plan

Run against `32f4dbe`.

| Gate | Command | Result |
|---|---|---|
| Type check | `bun run check` | 43 tasks successful |
| Type check — `scripts/` | `bunx --bun tsc -p scripts/tsconfig.json` | clean |
| Tests | `bun run test` | 44 tasks successful |
| Browser tier | `bun run test:browser` | 6 successful, 6 total — `@shared/ui` 62, `@cire/invites` 49, `@cire/host` 13, `@pulse/web` 12, `@musubi/social` 9 |
| Lab stories | `bun run --cwd tools/lab test:run` | 13 pass — every story renders headless |
| Lab stories, by eye | real Chromium, all 14 stories, both themes | screenshotted and reviewed; this is what found the stylesheet leak |
| Lint | `bun run lint` | 0 errors, 1052 warnings |
| Lint ratchet | `bash scripts/guard-lint-warnings.sh` | matches ceiling; ratcheted **down** 1058 → 1052 |
| Format | `bun run fmt:check` | clean, 1572 files |
| Changesets | `bash scripts/validate-changesets.sh` | green |
| Changeset required | `git diff --name-only origin/main...HEAD \| bash scripts/changeset-required.sh` | `required`, and present |
| Lockfile | `bun install --frozen-lockfile` | no changes — `bun.lock` hand-spliced for the split and for `@tools/lab`'s two new deps, never regenerated |
| Wikilinks | every `[[target]]` under `wiki/` resolved against the page set | 14 unresolved, all pre-existing false positives (`[[d1_databases]]` and friends are TOML table names) |
| Bundle — `@cire/invites` | `bun run --cwd cire/invites build` | 166,510 bytes gzip, budget 175,000 |
| Bundle — `@cire/landing` | `bun run --cwd cire/landing build` | 164,847 bytes gzip; threshold followed the measurement **down** to 176,547 |
| Bundle — `@cire/host`, `@cire/vendor` | `bun run --cwd cire/host build` | **not run locally** — the Astro font provider cannot reach `fonts.google.com` from this environment. Green in CI. |
| CI | — | green on `32f4dbe` (9/9) |

By hand, and honestly unverified here:

- **The two cire bundle budgets were not measured locally.** Only `@cire/invites` and `@cire/landing` were. CI's `guard-bundle-size.sh --all` covers all four and is green, so this is a gap in my own checking rather than in the evidence.
- **The bundle figures above predate the split** and were not re-measured after it. The split moves modules between packages without adding or removing any, so no app's graph should change — but that is a prediction, and CI's guard is what actually tests it.
- **`ModuleSidebar`'s bench lost the portal's shapes and spacing** when its stylesheet import was removed. Layout, placement, timing and interaction — what that bench exists for — are unaffected, but anyone who used it to eyeball portal spacing should know. A scoped way to give it back is worth an issue if that matters.
- **The toast's paint order over an open sheet is reasoned, not pixel-checked** — see the Decision above. A reviewer with a browser should open an invite, submit an RSVP and confirm the toast is visible over the sheet during the dwell.
- **`bun run test:scripts` cannot be trusted from my environment.** It fails on `scripts/tests/cire-dev-db-guard.cli.test.ts`, which asserts a Bun TOML parse message that changed — it fails identically on `origin/main`'s copy of that file, and this container runs Bun 1.3.11 while CI pins 1.4.2 via `.bun-version`. CI's `script-tests` job is green.
- **The browser tier needed an environment override here**, not in CI: this container ships Chromium 1194 under `$PLAYWRIGHT_BROWSERS_PATH` and the pinned Playwright wants a different build, so the runs above set `VITEST_BROWSER_EXECUTABLE_PATH`. That escape hatch is the one the vitest config already documents; CI installs the matching browser and leaves it unset.
- Visual review of the eleven contrast fixes: the ratios are arithmetic, but whether cire's primary button still looks like cire's primary button is a judgement.

https://claude.ai/code/session_018T5nsTR2iAhj8TkYotKuZC

<details><summary>Session metrics — $444.99 · 688.2M tok · complexity 5 · +10230/-5341 source</summary>

| | |
|---|---|
| Cost (API-equivalent) | $444.99 |
| Tokens | 688.2M — out 1.2M · cache-w 6.7M · cache-r 680.4M (99%) |
| Models | claude-opus-5 (94%), claude-fable-5-1 (6%), &lt;synthetic&gt; (0%) |
| Active time | 11h 10m over 1 session(s) |
| Declared complexity | 5 |
| Source diff | +10230/-5341 across 292 file(s), 23 package(s) |
| Turns | 61 (60 corrective) |
| Before first edit | 1.9M (0%) |
| Subagents | 2× attacker, 1× general-purpose, 2× implementer (8% of spend) |

<sub>Card: `.claude/metrics/claude-nifty-cerf-jrox2u.json` · phase `at-open` · [schema](../blob/main/wiki/observability/session-metrics.md)</sub>
</details>


chav-aniket:
## Adversarial review round — `0a8b35a`

A second model read the whole branch cold with one instruction: find where it derailed, and check the consent dialog hardest. Ten findings. Six are fixed here; the rest are recorded below with what was done instead.

### The consent dialog is safe, and the review says so on evidence

Every route still opens it (banner "Choose", the footer link, `/privacy`, `/terms`), a dismissal is still no decision, Reject is still exactly as reachable as Accept, and the store, the host arbitration and the reload-on-revoke are the same code they were. The focus trap is in fact **stronger** than the hand-rolled one it replaced, which listed no `select` or `textarea` in its `FOCUSABLE_SELECTOR` and could be tabbed out of.

But the review found two real problems *around* it, and one of them was mine.

### Fixed — a defect I shipped and no string assertion could see

`Modal`'s defaults are `base:`-prefixed — `:where(…)`, zero specificity — so that a **plain** caller utility wins. A `base:` one *ties*, and a tie resolves by Tailwind's stylesheet order: in `cire/invites`' built CSS, `.base\:max-w-osn-sm` is emitted at byte 42993 and `.base\:max-w-lg` at 42939, so the component beat its own caller.

Measured on the consent dialog before the fix — 480px instead of 512px, painted on `--osn-surface-raised` instead of the page ground, which put the panel on the same colour as the `bg-surface-raised/40` category rows inside it. The rows stopped reading as rows, on the one surface whose whole job is being legible.

`AnimatedModal` got this right and documented it. Nine call sites written after it did not. All nine fixed, with a browser test on the consent dialog's surface, width and row contrast, and the rule written up in `component-library.md`. **The durable guard is a new lint rule in the stacked PR #1062**, which found forty more sites.

### Fixed — a guarantee that quietly became false

`Z_LAYER.CONSENT (200) > MODAL (100)` used to put the consent **banner** over the details sheet. The sheet is a top-layer `<dialog>` now, so while one is open the banner is painted beneath it *and* inert. Both `z-index.ts` and `wiki/architecture/cire-consent.md` still claimed it sat "above everything".

Both now say what actually happens. The gap it leaves — an undecided guest has no consent affordance beside a third-party embed *inside* an open sheet — is #1061 with three options and a `needs:decision` label, rather than a paragraph nobody owns.

The **dialog** is unaffected and is in a stronger position than the number gave it: it is a `showModal()` dialog too, so opening it from inside a sheet makes it the blocking dialog and the sheet goes inert beneath it.

### Fixed — the rest

- **`TOAST < CONSENT` is a claim about `z-index`** and stops applying while `topLayer` has the toaster in the top layer. Said plainly in `Toaster`'s doc and `toast.md` instead of left as an invariant of the pixels.
- **The wiki under-delivered against the plan.** 130 lines about `Modal`'s exit choreography, and nothing about the token contract itself. New `wiki/architecture/design-tokens.md`; `@cire/ui`, `@shared/color` and `@shared/design-tokens` added to `monorepo-structure.md` and CLAUDE.md, where they existed in three commits and no page.
- **Reduced-motion coverage was lost** with the overlay it belonged to. `Modal` has it back, plus the `emulateMedia` browser command needed to write it. The 1ms clamp is load-bearing: the exit is *awaited*, so `none` would leave `getAnimations()` empty and take the close down a different path from every other engine.
- **Two comments narrated history** rather than stating what the code guarantees. Rewritten. The older ones elsewhere pre-date this branch and are not its to rewrite.

### Not fixed, and why

- **The toast is inert under an open sheet.** The review agreed this is acceptable as documented and suggested a simpler fix: raise the toast when the dwell timer closes the sheet rather than when the save lands, which would make it both hit-testable and announced. That is a product timing change to the RSVP confirmation, not a refactor — worth doing, worth doing deliberately, and it needs a number for the dwell that someone has looked at.
- **CSS Modules were never evaluated**, which the original brief asked for alongside StyleX. True. Most of the StyleX rejection carries over — six of ten frontends are Astro, and CSS Modules there would mean giving up the `data-[…]`, `group-` and container-query variants the layout is built on — but nobody wrote that down and it is not the same argument. Say the word and I will evaluate it properly rather than assert it.

### One test that was fragile by construction

Two Escape tests used `requestClose()`, which goes through the close-watcher budget: without user activation a page gets one free watcher, so it quietly stops closing anything once a suite has opened a few dialogs. The failure moved between two files across runs. Both now press Escape for real through `userEvent.keyboard`, which is the better test anyway. Two consecutive full browser runs, 49/49.

### What the review found nothing wrong with

`Modal` itself — the `close` listener wired via `addEventListener` rather than a JSX prop, the jsdom and SSR guards on `showModal`/`close`/`getAnimations`, the unmount that closes the element so the page is never left inert, the reopen-during-exit token. The `AddToCalendar` fix. The `@osn/ui` / `@cire/ui` split. Guard discipline — changesets, the version-less split, the ratchet, the bundle budgets, the lockfile. And the consent semantics, which were never at risk.

---

chav-aniket:
## CI was red, and it was mine — fixed in `a7b18cd`

**Type Check** failed on all three heads. The cause is a gate I was not running locally:

```
bunx --bun tsc -p scripts/tsconfig.json
scripts/codemod-scale.ts(60,33): error TS5097
scripts/tests/codemod-scale.test.ts(13,31): error TS5097
scripts/tests/codemod-scale.test.ts(98,46): error TS5097
```

`scripts/` is not a workspace, so `turbo check` never reaches it and it gets a CI step of its own. I ran `bun run check` and not that. Three imports were written with a `.ts` extension where `moduleResolution: bundler` wants none — every other script-to-script import in the tree is already extensionless. Dropped, `tsc -p scripts/tsconfig.json` now exits clean, and `scripts/tests/codemod-scale.test.ts` still passes 12/12.

Worth noting for the reviewer: `scripts/codemod-scale.ts` breaks the invariant `scripts/tsconfig.json`'s own comment states — that everything under `scripts/` imports only `bun:test`, Bun and Node built-ins, and each other. It reaches into `shared/design-tokens` for `SCALE_MIGRATION`, deliberately, so the migration table has exactly one home. That is why the relative path is long, and it is the reason this file was the one that tripped the rule.

**`@musubi/social#test:browser`** also failed, on `7d0c6dc` only. It died before any test body ran — `Tests: no tests`, one navigation timeout, with `Port 63315 is in use, trying another one…` repeated as six browser projects started concurrently under turbo. It passed unchanged on the next head (`0a8b35a`), so it is already resolved and I have not spent a re-run on it.

**`CI`** is the aggregate gate and was only reporting the above.

Also in this push: two `@vitest/browser/context` imports become `vitest/browser`. Vitest printed a `DEPRECATED` line for each on every browser run, and the module stops working in the next major.

### One thing I cannot verify here

`bun run test:scripts` fails in this environment on `scripts/tests/cire-dev-db-guard.cli.test.ts` — it asserts Bun's TOML parse error contains `"TOML"`, and the message is now `Expected t_equal but found is`. **It is not this branch's**: it fails identically with `origin/main`'s copy of that file, and this container runs Bun 1.3.11 while CI pins 1.4.2 via `.bun-version`. The `script-tests` job is green in CI, so this is a local-environment artifact and I have left it alone rather than "fixing" a test against the wrong Bun.

---

chav-aniket:
## The browser tier runs serially now — `90c1135`

CI went green on `faef803`, but the browser job had failed on three of the six runs this branch has had, and the pattern is worth naming rather than re-running past.

| Head | Failed |
|---|---|
| `c4ef945` | `Build & Test` |
| `7d0c6dc` | `@musubi/social#test:browser` |
| `da7e534` | `@osn/ui#test:browser` |

Each time the same shape: three rounds of `Port 63315 is in use, trying another one...`, then one project dies with `Tests: no tests, Errors: 1` on a navigation that never completed — while every test that *did* run passed. **A failure that lands on a different package each time is contention, not a test.**

And it is this branch's doing. `test:browser` ran two projects before; `@osn/ui`, `@musubi/social` and `@pulse/web` each gained one here, so six Vite servers and six Chromiums now start at once on a four-core runner.

`--concurrency=1`, which is exactly what `test:d1` already carries for the neighbouring reason that it boots workerd. The flag goes in the root script because turbo has no per-task setting; the reason goes in `turbo.json` beside the task, where someone editing it will read it.

It costs wall-clock — six projects at 15–60s each, serial, instead of about a minute overlapped. That is the right trade. An intermittent red that moves between packages teaches everyone to re-run rather than to look, and the browser tier is the only place several of this branch's guarantees can be checked at all.

Verified locally: `bun run test:browser` → `Tasks: 6 successful, 6 total`.

---

PR #1062, opened 2026-09-17
Design system phase 6: make the contract enforceable, then enforce it

### LINKED ISSUE #652: Design system: one token contract, two component layers

### LINKED ISSUE #1049: Design system phase 6: enforce the contract with @shadcn/lint

chav-aniket:
## Summary

The last of #652's eight phases. #1060 made the token contract exist; this branch makes it **enforceable**, and then enforces it: `shadcn/no-restyle` and `shadcn/no-arbitrary-values` both go to `error` with every surface already at zero.

- **176 call sites cleared** — 158 `no-restyle` and 18 `no-arbitrary-values` — across cire, musubi, pulse and the lab, by naming the design rather than by suppressing the rule. No `oxlint-disable`, no widened `allow`, no class picked to slip past the grammar.
- **A further 43 came in with `main`.** Merging `main` brought feature work written while the rules were still `warn`; every one of those is fixed here too, on the same terms.
- **The migration found seven component gaps by being blocked on them**, four in `@shared/ui` and three in `@cire/ui`. Each was first met by dropping the call-site class, and each turned out to be a regression rather than surplus styling — see **Decisions**.
- **One of them was already breaking a gate.** `bun run --cwd tools/lab check` was red at the branch tip before this work started, from a `DialogClose` type regression that #1060 shipped.
- **The migration itself introduced a regression, and the browser tier caught it.** 22 pixel width caps in `@cire/invites` were moved onto the rem spacing scale, which widened every desktop column by 6.25%. Fixed in `c52edda` — see **Decisions**.
- **Two `before:content-['•']` sites became real `<ul class="list-disc">` lists** rather than a rule exemption — an accessibility fix, not a neutral swap.
- **`.oxlintrc.restyle.json` is deleted.** Its configuration is now the real one in `oxlintrc.json`, so keeping it would be a second copy to drift.

## Workspaces affected

`@shared/ui`, `@osn/auth-ui`, `@musubi/social`, `@musubi/landing`, `@pulse/web`, `@pulse/landing`, `@tools/lab`, `@cire/ui`, `@cire/host`, `@cire/invites`, `@cire/landing`.

Five new changesets on top of the branch's existing ones, split so the version-less `@cire/*` never share a file with a versioned package.

## Issues

**Closes**

| Issue | What |
|---|---|
| #1049 | Design system phase 6: adopt `@shadcn/lint` and take its rules to `error` |
| #652 | Design system epic — its eighth and last sub-issue is #1049 |

Closes #1049
Closes #652

**Opened**

| Issue | What |
|---|---|
| #1117 | `@pulse/landing` writes 53 literal type values across 23 distinct sizes in seven `.astro` files, which oxlint cannot see |
| #1118 | `shadcn/no-restyle`'s layout grammar rejects any `max-w-*` outside Tailwind's built-in steps, so there is no class form of a pixel cap on a library component |
| #1119 | The events drag grip may not meet the WCAG 2.5.8 24px minimum target, and a comment asserts a `py-2` that is not there |
| #1120 | `component-library.md` documents a named `@cire/ui` `Button` import against a default export |

### Four `@shared/ui` gaps the migration found by hitting them — `approach`

- **Issue** — four treatments call sites were spelling by hand had no variant to move to, so the first pass dropped the class and took the regression.
- **Why** — each was a real defect, not surplus styling. `AvatarFallback` has no size, so musubi's 64px settings avatar rendered its initials at **12px** once its `text-display` was dropped. `Button` has no quiet-destructive variant, so four ghost row actions became **filled red at rest**, which reads as an error state down a column. `DropdownMenuTrigger` was a bare Kobalte re-export with **no focus treatment at all**, so an avatar-as-trigger gave a keyboard user nothing and its contract forbade the call site fixing it. `DialogClose` had been given a concrete `ComponentProps<"button">`, which dropped Kobalte's polymorphic `as` from the **type** while leaving it working at runtime.
- **Solution** — `AvatarFallback` takes a `size` (two steps, not three); `Button` takes `ghostDanger`; `DropdownMenuTrigger` takes a `treatment`, `bare` by default so the `as={Button}` spelling is untouched; `DialogClose` and `DropdownMenuTrigger` are polymorphic again.
- **Rationale** — two independent agents hit the `AvatarFallback` wall separately and took the same lossy answer, which is what made it a gap rather than a preference. `Button`'s case has a precedent in the repo that settles it: `cire/ui`'s own `bareDanger` carries a comment recording that **dropping exactly this treatment had already been tried and reverted** — "five of these were `text-text-muted hover:text-error`, and folding them into `bare` left a delete action that no longer signals anything." `AvatarFallback` sizing is one of the two items #1060 deferred on design grounds as "a non-constant ratio"; two steps keyed to the two jobs the call sites actually distinguish is the answer, and an open scale would have been three names guessing at sizes nothing needs. The other deferral, `Badge`'s eyebrow treatment, is taken up here too.

### A px cap is not a rem cap in the one app that resizes its root — `defect`

- **Issue** — the migration converted 22 `max-w-[Npx]` caps in `@cire/invites` onto Tailwind's spacing scale (`max-w-[640px]` → `max-w-160`). The earlier changesets called these exact swaps.
- **Why** — they are exact only at a 16px root, and `@cire/invites` is the **only** app in the repo that changes its root: `src/styles/global.css` steps it 16px → 17px at 1024px. The scale is rem, so every converted cap became **6.25% wider on every desktop viewport**. The classic pack's events column went 640px → 680px, gala's 960px → 1020px, on the live guest site.
- **Solution** — a named `--container-column-*` scale in the app's own `@theme`, in px, with the reason on the block. Each source was checked before conversion: the sites that came from `34rem` had already become contract `max-w-ui-*` tokens and are untouched, as is the one that came from `26rem`.
- **Rationale** — found by `EventCard.actions.browser.test.tsx`, whose wrapped case measures the card past the 1024px step; at 680px the row stopped wrapping and the assertion named that cause. It had been invisible until then because the test's own harness carried the same swap, so it had been widened to match production and was no longer measuring the real cap. Two lessons worth the space: a px→rem conversion is only value-preserving where the root is fixed, and a measurement harness that tracks the thing it measures has stopped being a measurement.

### The `DialogClose` regression was already red, and only one package could see it — `defect`

- **Issue** — `bun run --cwd tools/lab check` failed at `8cbb4a0`, before any of this branch's remaining work.
- **Why** — `DialogClose`'s own doc comment names `<DialogClose as={Button}>Cancel</DialogClose>` as the pattern its `bare` treatment exists to serve, and `tools/lab` is the **only** consumer of that pattern repo-wide. So the type regression broke exactly one package's typecheck and nothing else, and the contract and the behaviour disagreed silently everywhere else.
- **Solution** — carry Kobalte's `PolymorphicProps` on both wrappers. Six further lab errors were the radius maps not catching up with the contract's new `--ui-radius-sheet`, fixed alongside.
- **Rationale** — the story was not deleted to make the gate green. It is the only demonstration of a documented API, and deleting it would have hidden the defect rather than found it.

### Merging a squash-merged parent — `approach`

- **Issue** — #1060 was **squash**-merged into `main`. This branch carries its original unsquashed commits, so git's merge base predates it and cannot tell that identical content on both sides is the same change. `git merge origin/main` produced **108 conflicts**.
- **Why** — the naive readings are both wrong. Taking `main` wholesale reverts the migration; taking this branch wholesale drops four merged PRs of cire feature work.
- **Solution** — the conflicts were sorted by whether `main` had actually changed the file *after* the squash commit. **75 were pure squash artifact** — this branch's side is strictly newer, so taking it loses nothing. **7 `package.json`** took `main`'s version bumps and new `@cire/dietary` deps; none of this branch's commits touched a `package.json`. **`shared/ui/src/ui/popover.tsx`** was a genuine both-sides change — `main` added a top-layer portal fix for a popover opened inside a modal dialog, this branch added the `padding` variant, and both are kept. The remaining **21 cire files** were resolved per hunk: `main`'s feature wins wherever it deleted the markup the migration had converted, the migration wins wherever `main`'s form is a restyle the rules now reject.
- **Rationale** — the tempting shortcut is `git merge ac41e37 -s ours`, which declares the squash already merged. It is wrong here: the parent branch tip and the squash commit differ by nine `.claude/` files from a PR that landed between them, so it would have silently dropped them. Checking that took one `git diff` and is the reason it was not used.

### `bun.lock` could not be taken from either side — `defect`

- **Issue** — taking `main`'s lockfile left the tree un-installable: `main` has no `@shadcn/lint`, which this branch's root `package.json` requires. `bun install --frozen-lockfile` failed outright.
- **Why** — neither side's lockfile is correct for the merged `package.json` set. `main`'s lacks the `@shadcn/lint` subtree (107 entries); this branch's lacks `@cire/dietary`. Worse, the two resolve `oxc-parser` **differently** — `@shadcn/lint`'s optional dependency on it causes `@solidjs/start` to carry a nested copy — so no splice of one into the other passes. Two were attempted and both failed: a `patch(1)` of one delta onto the other applied at fuzzy offsets and corrupted the workspace block, and an order-preserving entry merge still left bun wanting changes.
- **Solution** — regenerated, then checked against the pruning `CLAUDE.md` warns about. The 28 entries that went are a nested `oxc-parser` subtree hoisting — the set includes its **linux** bindings, where a platform prune would have kept those and dropped only the others — and the `darwin` and `win32` counts now match `main`'s exactly, 59 and 51.
- **Rationale** — worth recording because "regenerate the lockfile" is the move this repo tells you not to make, and the check that makes it safe is a specific one: compare the platform-entry counts against a reference tree rather than eyeballing the diff size.

### The bullets became lists rather than an exemption — `approach`

- **Issue** — `ImportPanel.tsx` and `ChangePreview.tsx` rendered warning bullets as `before:content-['•']`, the two sites expected to need a rule exemption.
- **Why** — both are `<ul>`s laid out as `flex flex-col`. A flex item is blockified and loses its `::marker`, which is *why* the bullet was generated content.
- **Solution** — both are real lists now: `<ul class="… list-disc pl-5">` with bare `<li>`, so the platform draws the marker. `oxlintrc.json` was not touched for either.
- **Rationale** — an accessibility improvement rather than a neutral swap. Some screen readers announce generated content as content, and a list-styled `<ul>` keeps its list semantics.

### The warning ceiling took `main`'s number unchanged — `approach`

- **Issue** — `scripts/lint-warning-ceiling.txt` is a two-way ratchet: a count *below* the committed number fails too.
- **Why** — before the merge this branch was failing it at 1070, which is 1052 plus exactly the 18 `no-arbitrary-values` warnings the rule was emitting at `warn`. Clearing them returned the count to the 1052 already in the file, so no edit was needed. `main` has since ratcheted down to **1047**.
- **Solution** — take `main`'s 1047 and its comment verbatim. After the merge the guard reports `1047 lint warnings, matches the ceiling`.
- **Rationale** — the file's own instruction is to keep the branch and date current **on every re-baseline**. There was no re-baseline on either side of this, so rewriting that line would have read as a measurement that never happened.

### Arbitrary values become app tokens, not contract tokens — `approach`

- **Issue** — 18 sites hardcoding a value with no token to move to, plus 15 more that arrived with `main`.
- **Solution** — each becomes a named entry in **that app's own theme block**: `--tracking-eyebrow` and `--leading-display` in `@musubi/landing`; six type and tracking steps in `@pulse/landing`; `--container-hero` and `--spacing-accent-word` in `@pulse/web`; `--blur-scrim`, `--text-glyph`, `--drop-shadow-pin`, `--blur-ground-shadow`, two `--ui-radius-sheet` mappings and the `--container-column-*` scale across `@cire/*`.
- **Rationale** — the `ui-*` contract is library-facing, so an app's own hero eyebrow does not belong in it. Pulse's two hero font sizes are **named rather than snapped**: 9.6px and 10.4px both sit below the scale's smallest step, so rounding them up would change the eyebrow rather than tidy it. Every entry was verified against the **built CSS** rather than assumed, because a wrong `@theme` entry emits no rule at all — no error, no warning.

### Both rules were proved to fire before their zero was trusted — `approach`

- **Issue** — a rule at `error` reporting nothing is indistinguishable from a rule that is not running.
- **Solution** — a probe component carrying `bg-red-500`, `rounded-full` and `text-lg` on a `<Button>` plus `p-[13px]` and `text-[0.9rem]` on a `<span>` produced five errors and a non-zero exit; it was then removed and lint returned to exit 0. The probe was run **again** after the lockfile was regenerated, to prove the plugin still loads.
- **Rationale** — the same evidence every rule enabled in this series has had to produce. The probe's output also confirmed the new variants registered: the rule's own suggestion list reads `default, destructive, outline, secondary, ghost, ghostDanger, link`.

**Out of scope** — four follow-ups, filed rather than smuggled in: #1117, #1118, #1119, #1120. Details in each.

## Test plan

Run against `c52edda`.

| Gate | Command | Result |
|---|---|---|
| `no-restyle` | `oxlint -c oxlintrc.json .` | **0** (from 158, plus 43 that arrived with `main`) |
| `no-arbitrary-values` | `oxlint -c oxlintrc.json .` | **0** (from 18) |
| Both rules fire | probe component, then removed | 5 errors, exit 1; exit 0 after removal |
| Type check | `bun run check` | 44 tasks successful |
| Tests | `bun run test` | 45 tasks successful |
| Lint | `bun run lint` | **0 errors**, 1047 warnings |
| Lint ratchet | `bash scripts/guard-lint-warnings.sh` | matches ceiling at 1047 |
| Format | `bun run fmt:check` | clean, 1613 files |
| Changesets | `bash scripts/validate-changesets.sh` | green |
| Changeset required | `git diff --name-only origin/main...HEAD \| bash scripts/changeset-required.sh` | `required`, and present |
| Lockfile | `bun install --frozen-lockfile` | clean — see the `bun.lock` Decision |
| CI, all 8 checks | — | **green on `c52edda`**, Build & Test included |

By hand, and honestly unverified here:

- **The browser tier cannot run in this container** — its prebuilt Playwright is build 1194 and the pinned Playwright wants 1234. **CI runs it, and it is green on `c52edda`.** That matters more than usual on this branch: the browser tier is what caught the px/rem regression above, and it failed once on `25eb710` before the fix. Locally, every `.browser.test.tsx` covering a changed component was read and its assertions checked against the change by hand.
- **`@cire/ui` gained a `field` variant** during the merge resolution — a control that reads as a form field rather than an action, for the dietary-presets popover trigger `main` added. It is new design vocabulary introduced while resolving a conflict, which is not where design decisions normally get made, so it is the change here most worth a second opinion.
- **`AnimatedModal` carries its 480px cap as an inline `max-width`** rather than a class, because `no-restyle`'s grammar rejects any `max-w-*` outside Tailwind's built-in steps and every built-in step is rem. That is #1118, and the inline style should go back to a class once it is resolved.
- Whether the seven new variants look right is a judgement, not an assertion. The deliberate visual changes are enumerated per-tree in the commit messages; the largest are cire's four `tile` call sites unifying on one wash, and `@cire/host`'s collapsed section-menu trigger growing from 44px to ~51px.

https://claude.ai/code/session_013dETQp6EnsTQDH6jSicG1L


chav-aniket:
## Handoff — where the `shadcn/no-restyle` migration stands

Picking this up in a fresh session? Everything you need is below. Branch is `claude/nifty-cerf-lint` at `8cbb4a0`.

### What has landed

| Commit | What |
|---|---|
| `a2fd2cd` | `shadcn/no-unknown-classes` at `error`, plus the overlay animation utilities the dead `slide-*` classes were standing in for |
| `38d4c89`, `b6dafb6`, `de86ba6`, `841c187` | 284 call sites cleared across musubi, pulse, cire |
| `3f3e089` | **`@shared/ui` variants** for the treatments call sites were spelling by hand, plus `--ui-radius-sheet` |
| `7937c66` | `Modal`'s plane is its own `surface` prop; `wiki/architecture/component-library.md` and `design-tokens.md` updated |
| `8cbb4a0` | `.oxlintrc.restyle.json`, the measuring config |

### The remaining work

**158 `shadcn/no-restyle` sites** and **18 `shadcn/no-arbitrary-values` sites**. Both rules go to `error` in `oxlintrc.json` once the counts are zero. Measure:

```bash
./node_modules/.bin/oxlint -c .oxlintrc.restyle.json . 2>&1 | grep -c "shadcn(no-restyle)"     # 158
./node_modules/.bin/oxlint -c oxlintrc.json . 2>&1 | grep -c "no-arbitrary-values"             # 18
```

| Package | `no-restyle` |
|---|---|
| `pulse/web` | 51 |
| `cire/host` | 43 |
| `musubi/social` | 42 |
| `cire/invites` | 13 |
| `cire/landing` | 7 |
| `osn/auth-ui` | 1 |
| `tools/lab` | 1 |

`.oxlintrc.restyle.json` carries the `allow` and `contracts` the real config will take: `allow: ["layout"]` plus a contract letting `gap-*` through on `Card`, `CardContent`, `CardFooter`, `CardHeader`, `Modal`, `DialogFooter`, `DialogHeader`, `PopoverTrigger` and `DropdownMenuTrigger` — containers whose children are the caller's. That accounts for the 175 → 158 difference against a bare run.

### The new `@shared/ui` API to migrate onto

Do not change `shared/ui/` or `shared/design-tokens/` — they are done.

| Component | Prop | Values |
|---|---|---|
| `Modal` | `presentation` / `surface` | `centred`\|`sheet`\|`drawer` / `raised`\|`surface`\|`ground` |
| `DialogContent` | `presentation` | `centred`\|`sheet` |
| `DialogClose` | `treatment` | `bare`\|`glyph` |
| `Badge` | `treatment` | `default`\|`eyebrow` |
| `Label` | `tone` / `size` | `default`\|`muted` / `sm`\|`md`\|`lg` |
| `Avatar` | `ring` | `none`\|`surface` |
| `Card` | `variant` / `elevation` / `radius` | `default`\|`muted`\|`dashed` / `flat`\|`raised` / `md`\|`lg` |
| `PopoverContent` | `padding` | `default`\|`tight`\|`none` |
| `DropdownMenuLabel` | `tone` | `heading`\|`identity` |
| `TabsTrigger` | `size` | `sm`\|`md` |
| `Td` | `code`, `indent` | booleans, beside the existing `numeric` |
| `Input` | `face` | `text`\|`code` |

`Modal`'s `surface` defaults with the presentation — `raised` for `centred`, `surface` for `sheet` and `drawer` — so a call site usually passes nothing.

### Four rules, and the reason each exists

1. **Never suppress.** No `oxlint-disable`, no class picked to slip past the grammar, no widening of the `allow` list. A rule at `error` reporting nothing is indistinguishable from a rule that is not running — every rule enabled so far was proved to fire on a deliberate violation before its zero was trusted.
2. **A variant is a design concept with a name and a reason.** Match the register of `cire/ui/src/button.tsx` and `shared/ui/src/ui/modal.tsx`: say what it is *for*, and why it is one choice rather than three classes. Never `custom`, `alt` or `variant2`.
3. **`base:` inside a component, plain utilities at a call site.** `base:` is `:where(&)`, zero specificity, so a component default loses to a caller's plain utility by design and *ties* with a caller's `base:` one. Responsive form is `md:base:x` / `max-md:base:x` — verified to compile and to be emitted after the unprefixed `base:` rules.
4. **No comment narrates history.** `house/no-tracker-ref-in-comment` is at `error`.

### Two decisions already taken, so they are not re-litigated

- **`ConsentPreferences` keeps `--color-bg`.** `cire/invites/tests/components/consent/consent-dialog-surface.browser.test.tsx` asserts that panel paints on the page ground and specifically not on a raised surface, because its own category rows *are* raised surfaces and the panel behind them has to sit below them. Use `presentation="sheet" surface="ground"`. It does move from an `sm:` breakpoint to the shared `md:` one — a sheet up to 768px rather than 640px.
- **`no-raw-colors` cannot reach `error` at any effort**, and is not part of this. 844 of its 905 reports are the plugin reading custom font-size tokens (`text-ui-sm`, musubi's `text-body`/`text-meta`) as colours. Not configurable.

### The 18 arbitrary values, by package

```
cire/invites  AnimatedModal.tsx:117   rounded-t-[1.75rem]   → --ui-radius-sheet: 1.75rem in the app's mapping
cire/landing  demo/DemoModal.tsx:43   rounded-t-[1.75rem]   → same
cire/host     ImportPanel.tsx:469     before:content-['•']  → a glyph, not a size; likely an exemption
cire/host     ChangePreview.tsx:133   before:content-['•']  → same
cire/invites  MapPreview.tsx:158,165  drop-shadow-[…], blur-[1px]
cire/host     CommandPalette.tsx:226  backdrop-blur-[2px]
cire/host     ModuleSidebar.tsx:228   text-[0.95em]
musubi/landing ConnectionsHero.tsx:186,189  tracking-[0.26em], leading-[1.05]
pulse/landing PulseHero.tsx:108–136   text-[0.6rem], text-[0.65rem], 3× tracking-[…], leading-[1.05]
pulse/web     explore/ExploreNav.tsx:201,210  max-w-[16ch], mr-[0.2em]
```

A value that is a design decision moves to a **named entry in that app's own theme block** — not to the `ui-*` contract, which is library-facing. `PulseHero`'s two sizes sit below the shared scale's smallest step (`0.7rem`), so snapping them is a visible change on a hero eyebrow; naming them in pulse's own theme is the honest fix.

### Closing steps, in order

1. Clear both counts to zero.
2. `oxlintrc.json`: `shadcn/no-restyle` to `error` with the `allow` + `contracts` from `.oxlintrc.restyle.json`, and `shadcn/no-arbitrary-values` from `warn` to `error`.
3. Prove each newly-`error` rule fires on a deliberate violation, then revert the probe.
4. Re-baseline `scripts/lint-warning-ceiling.txt`. It is a **two-way** ratchet — a count below the ceiling fails too. The current 1070 is 1052 plus exactly the 18 `no-arbitrary-values` warnings; clearing them takes it back to 1052 and the file must follow, with the branch and date in its trailing comment.
5. Changesets per package, never mixing version-less `@cire/*` with a versioned package in one file. Check with `git diff --name-only origin/main...HEAD | bash scripts/changeset-required.sh`.
6. Update this PR body, take it out of draft, add `Closes #1049` and `Closes #652`.

### One gate that cannot run here

`bun run test:browser`. The container's prebuilt Playwright is build **1194** and the pinned Playwright wants **1234**; `npx playwright install` is not available in this environment. CI runs it. Say so rather than reporting it green.

---

